PR #2, opened 2023-05-02
Add relations to metadata

sed-i:


metadata.yaml
```diff
@@ -10,8 +10,53 @@ description: |
 requires:
   mimir-worker:
     interface: mimir_worker
+    description: |
+      Coordinator forwards operations data to the workers:
+      - s3 storage information
```
PietroPasotti, line 15: and hash ring and config, but yes

PietroPasotti: also, do we need to forward loki/tempo endpoints?

PietroPasotti: [needs discussion]: above

sed-i: Done

metadata.yaml
```diff
@@ -10,8 +10,55 @@ description: |
...
+      Forward workers' built-in logging rules to the ruler.
+
+  grafana-dashboards-consumer:
+    interface: grafana_dashboard
+    description: |
+      Forward workers' built-in dashboards to grafana.
+
+  ingress:
+    interface: ingress_per_unit
+    limit: 1
```
Abuelodelanada, line 45:
```suggestion
  ingress:
    interface: ingress
    limit: 1
```

Abuelodelanada:
Going with option 1. Merging

Issue #4, opened 2023-05-03
distributed mimir deployment scenario


lucabello:
The deployment scenario should follow one of the following options.

## Option 3: gagent as gateway towards cos-lite

```mermaid
graph LR

subgraph mimir
worker1 ---|mimir-worker| coordinator
worker2 ---|mimir-worker| coordinator
worker3 ---|mimir-worker| coordinator
coordinator ---|logging| gagent
coordinator ---|scrape| gagent
coordinator ---|remote-write| gagent
coordinator ---|dashboards| gagent
coordinator ---|trace| gagent

end

subgraph cos-lite
gagent ---|logging| loki
gagent ---|remote-write| prometheus
gagent ---|dashboards| grafana
gagent ---|trace| tempo

end
```

Dashboards are owned by the coordinator, as they are "per-app".
We need the grafana-agent to scrape the Mimir workers, because they are not able to remote write themselves.

## Option 4: gagent inside the worker charm

Another option is to have grafana-agent deployed in each worker as a sidecar.

```mermaid
graph LR

subgraph mimir
worker1+gagent ---|mimir-worker| coordinator
worker2+gagent ---|mimir-worker| coordinator
worker3+gagent ---|mimir-worker| coordinator


end

subgraph cos-lite
coordinator ---|logging| loki
coordinator ---|remote-write| prometheus
coordinator ---|dashboards| grafana
coordinator ---|trace| tempo

end
```

In this scenario, the coordinator owns both dashboards and alert rules. It will need to take care of duplicating the alert rules per each worker by adding their topology information. Self-monitoring happens:
* internally, with nginx providing the **remote_write** endpoint of the appropriate Mimir worker;
* externally, with the coordinator providing the Loki/Prometheus URLs so that the workers can remote write directly to them.

During an upgrade of the workers, there might be a temporary inconsistency between alert rules and metric names (i.e., if the metric names change). However, it might be something we don't mind as it simplifies the architecture considerably.

lucabello:
We decided to pivot to something similar to the **Option 5** presented in the comment above.

However, the worker-coordinator organization we agreed upon is quite different; the new one is presented below.

## Option 6: gagent in coordinator and component-specific relations

```mermaid
graph LR

subgraph mimir
worker:::charm --- compactor:::relation --- coordinator+gagent:::charm
worker --- distributor:::relation --- coordinator+gagent
worker --- ingester:::relation --- coordinator+gagent
[...]

PR #5, opened 2023-05-04
Option 6

sed-i:
This PR implements option 6 from #4.

TODO:
- [x] Update metadata to reflect option 6
- [x] CI + minimal tests
- [x] fetch-lib prometheus_remote_write (the coordinator, not the worker, owns all rule files) + stubs for:
  - gagent pushes workers' metrics back to mimir, and, as a backup, to an external prometheus
  - forward metric rules to the ruler
  - forward metric rules to the external prometheus
  - gagent does not forward its own metrics
- [x] fetch-lib grafana_dashboard (the coordinator, not the worker, owns all dashboards) + stubs for:
  - forward dashboards to grafana
- [x] fetch-lib loki_push_api (the coordinator, not the worker, owns all rule files) + stubs for:
  - Need to obtain Loki's endpoint URL
  - forward log alert rules to loki

```mermaid
graph LR

subgraph mimir
worker:::charm --- compactor:::relation --- coordinator+gagent:::charm
worker --- distributor:::relation --- coordinator+gagent
worker --- ingester:::relation --- coordinator+gagent
worker --- querier:::relation --- coordinator+gagent
worker --- query-frontend:::relation --- coordinator+gagent
worker --- store-gateway:::relation --- coordinator+gagent
worker --- alertamanger:::relation --- coordinator+gagent
worker --- ruler:::relation --- coordinator+gagent

classDef charm fill:#f70,stroke:#000,stroke-width:3px,color:#fff
classDef relation fill:#fbb,stroke:#000,stroke-width:2px,color:#000,stroke-dasharray: 5 5


end

subgraph cos-lite
coordinator+gagent --- logging:::relation --- loki:::charm
coordinator+gagent --- remote-write:::relation --- prometheus:::charm
coordinator+gagent --- dashboards:::relation --- grafana:::charm
coordinator+gagent --- trace:::relation --- tempo:::charm

end
```


src/charm.py
```diff
@@ -66,13 +98,29 @@ def _on_config_changed(self, event):
...
-            relation.data[self.app]['s3_storage'] = self._s3_storage
+            relation.data[self.app]["config"] = json.dumps(dict(self.model.config))
+            relation.data[self.app]["hash_ring"] = json.dumps(hash_ring)
+            relation.data[self.app]["s3_storage"] = json.dumps(self._s3_storage)
+
+    def _remote_write_endpoints_changed(self, _):
+        # TODO Update grafana-agent config file with the new external prometheus's endpoint
+        pass
+
+    def _on_ingester_changed(self, _):
```
rbarry82, line 113: Nominally, the endpoint will always be proxied through the nginx here, right? So the grafana-agent remote-write target _should_ generally be the coordinator itself.

sed-i: Not sure: if we scale up/down, would we need to add/remove grafana agent targets?

rbarry82: For targets, we definitely would, but that applies to.. basically every one, not just the ingester. It would basically be the same list as the hash ring memberlist (which is more or less the set of all units).

Remote write would still be the same proxied endpoint.

sed-i: Replaced with a comment in the next commit.

src/mimir_coordinator.py
```diff
@@ -37,37 +42,39 @@
...
 
-    def roles(self) -> Counter[str]:
+    def roles(self) -> Counter:
         """Gather the roles from the mimir_worker relations and count them."""
         roles = Counter()
 
         for relation in self.relations:
             if not self._relation_data_valid(relation):
-                logger.error(f'Invalid relation data in {relation!r}')
+                logger.error(f"Invalid relation data in {relation!r}")
```
Abuelodelanada, line 71: Not so critical, but:

```suggestion
                logger.error("Invalid relation data in %s", relation)
``

**Review by Abuelodelanada — APPROVED** [review]

File: metadata.yaml [hunk 3]
```diff
@@ -12,23 +12,61 @@ description: |
 containers:
   nginx:
     resource: nginx-image
+  agent:
+    resource: agent-image
+    mounts:
+      - storage: data
+        location: /tmp/agent/data
```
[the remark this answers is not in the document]
sed-i: I think I agree, but I copied it from the [gagent charm](https://github.com/canonical/grafana-agent-k8s-operator/blob/73543ef892912848c3babf062cbd17f8f7e20730/k8s_metadata.yaml#L31).

src/mimir_coordinator.py
```diff
@@ -37,37 +42,39 @@
...
+    """Mimir coordinator."""
+
     def __init__(self, relations: List[Relation]):
         self.relations = relations
 
     def is_coherent(self):
         """Return True if the roles list makes up a coherent mimir deployment."""
         roles = self.roles()
-        return set(roles).issubset(MINIMAL_DEPLOYMENT)
+        return roles and set(roles).issubset(MINIMAL_DEPLOYMENT)
```
PietroPasotti, line 58: [praise] heh, good point.

src/mimir_coordinator.py
```diff
@@ -37,37 +42,39 @@
...
 
-    def is_recommended(self):
+    def is_recommended(self) -> bool:
+        """Return True if is a subset of the minimal deployment."""
         roles = self.roles()
-        return set(roles).issubset(MINIMAL_DEPLOYMENT)
+        return bool(roles) and set(roles).issubset(MINIMAL_DEPLOYMENT)
```
PietroPasotti, line 63:
```suggestion
        return roles and set(roles).issubset(MINIMAL_DEPLOYMENT)
```

src/mimir_coordinator.py
```diff
@@ -37,37 +42,39 @@
...
 
-    def roles(self) -> Counter[str]:
+    def roles(self) -> Counter:
         """Gather the roles from the mimir_worker relations and count them."""
         roles = Counter()
 
         for relation in self.relations:
             if not self._relation_data_valid(relation):
-                logger.error(f'Invalid relation data in {relation!r}')
+                logger.error("Invalid relation data in %s", relation)
```
PietroPasotti, line 71: [disagree] I disagree. I like f-strings

src/mimir_coordinator.py
```diff
@@ -37,37 +42,39 @@
...
                 continue
 
             try:
-                raw_roles = json.loads(relation.data[relation.app]['roles'])
+                raw_roles = json.loads(relation.data[relation.app]["roles"])  # type: ignore
```
PietroPasotti, line 75: [suggestion] Is this not better solved with a `typing.cast`? We know what it should be, don't we?

PietroPasotti (approved):
no major blockers, just a couple of thoughts/nits

src/mimir_coordinator.py
PietroPasotti: actually, my bad, we're doing this all wrong. It should be issuperset!

PietroPasotti: or `MINIMAL_DEPLOYMENT.issubset(roles)` and then we don't need the `roles and` anymore.

PR #7, opened 2023-05-15
Issubset issuperset

PietroPasotti:
several fixes in coordinator logic and some fancy scenario tests to verify the coherent/recommended code

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -108,5 +108,5 @@ commands =
            -s \
            --tb native \
            --log-cli-level=INFO \
-           --ignore={[vars]tst_path}integration \
+           {[vars]tst_path}integration \
            {posargs}


lucabello (approved):

src/mimir_coordinator.py
```diff
@@ -55,14 +63,21 @@ def __init__(self, relations: List[Relation]):
...
+
+        I.E. If all required roles are assigned, and each role has the recommended amount of units.
+        """
         roles = self.roles()
-        return bool(roles) and set(roles).issubset(MINIMAL_DEPLOYMENT)
+        # python>=3.11 would support roles >= RECOMMENDED_DEPLOYMENT
+        for role, min_n in RECOMMENDED_DEPLOYMENT.items():
+            if roles.get(role, 0) < min_n:
+                return False
+        return True
```
dstathis, line 78: Suggestion: Can we not use issuperset here?

dstathis (approved):

src/mimir_coordinator.py
PietroPasotti: is_consistent does use supersets because there all we need to check is that all required roles are assigned (regardless of how many units are assigned to each role)

PR #9, opened 2023-05-25
Decouple workload, configuration management from charm.py

sed-i:
In this PR a few parts of operating grafana agent have been decoupled from `charm.py`.
Something similar is expected to follow in a future PR for operating nginx.

- Split out "WorkloadManager" and "Config".
- Set up stage for the config to be lazily evaluated, in an attempt to avoid code ordering issues.
- New "compound status" handling approach:
  - Charm components report back via a callback
  - Each component, and then the charm, are responsible for calculating the "total" status for themselves

Apart from that, most of the code was copy-pasted from grafana agent (k8s operator).

### DIFF
--- a/src/agent_workload.py
+++ b/src/agent_workload.py
@@ -1,4 +1,4 @@
-"""Workload manager for grafna agent."""
+"""Workload manager for grafana agent."""
 
 import logging
 import pathlib
--- a/src/agent_workload.py
+++ b/src/agent_workload.py
@@ -9,8 +9,8 @@ import yaml
 from ops.framework import Object
 from ops.model import ActiveStatus, StatusBase, UnknownStatus, WaitingStatus
 from ops.pebble import APIError, PathError
-from yaml.parser import ParserError
 from yaml.constructor import ConstructorError
+from yaml.parser import ParserError
 
 logger = logging.getLogger(__name__)
 
--- a/src/agent_workload.py
+++ b/src/agent_workload.py
@@ -10,6 +10,7 @@ from ops.framework import Object
 from ops.model import ActiveStatus, StatusBase, UnknownStatus, WaitingStatus
 from ops.pebble import APIError, PathError
 from yaml.parser import ParserError
+from yaml.constructor import ConstructorError
 
 logger = logging.getLogger(__name__)
 
@@ -162,8 +163,10 @@ class WorkloadManager(Object):
         assert config is not None
 
         try:
-            old_config = yaml.safe_load(self.read_file(self.CONFIG_PATH))
-        except (FileNotFoundError, PathError, ParserError):
+            old_config = self.read_file(self.CONFIG_PATH)
+            logger.info("Old config: %s", old_config)
+            old_config = yaml.safe_load(old_config)
+        except (FileNotFoundError, PathError, ParserError, ConstructorError):
             # The file does not yet exist?
             old_config = None
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -34,11 +34,10 @@ class MimirCoordinatorK8SOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
         self.grafana_agent_workload = WorkloadManager(
-            self, container_name="agent", config_getter=lambda: {}
-        )
-        self.framework.observe(
-            self.grafana_agent_workload.on.status_changed,  # pyright: ignore
-            self._update_unit_status,
+            self,
+            container_name="agent",
+            config_getter=lambda: {},  # TODO
+            status_changed_callback=self._update_unit_status,
         )
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
--- a/src/charm.py
+++ b/src/charm.py
@@ -81,7 +81,7 @@ class MimirCoordinatorK8SOperatorCharm(CharmBase):
                 insecure_skip_verify=True,
                 http_listen_port=3500,
                 grpc_listen_port=3600,
-            ).build,  # TODO figure out what to do about potential code ordering problem
+            ).build(),  # TODO figure out what to do about potential code ordering problem
             status_changed_callback=self._update_unit_status,
         )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -13,8 +13,11 @@ import json
 import logging
 from typing import List
 
+from agent_config import Config
+from agent_workload import WorkloadManager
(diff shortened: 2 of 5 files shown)


src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+                    "command": f"/bin/agent {self._cli_args()}",
+                    "startup": "enabled",
+                },
+            },
+        }
+        self._container.add_layer(self._service_name, pebble_layer, combine=True)
+        self._container.autostart()
+
+        if version := self.version:
+            self._unit.set_workload_version(version)
```
dstathis, line 109: issue: I'm not sure the version we display should be the grafana-agent version. This is a Mimir charm after all. Not sure what version we *should* display but I don't think it should be this.

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+            # Nothing changed, possibly new installation. Move on.
+            self.status.config = ActiveStatus()
+            return
+
+        try:
+            self.write_file(self.CONFIG_PATH, yaml.dump(config))
+        except APIError as e:
+            logger.warning(str(e))
+            self.status.config = WaitingStatus(str(e))
+            return
```
dstathis, line 183: question: Should the `write_file` be in an else block? Seems like it shouldn't be needed if the config file did not change.

question: Should we restart or reload grafana-agent after writing the new config file?

dstathis, line 183: issue: I do not think waiting status makes sense. We are not waiting on a related app. In fact, this error is not resolvable at all, so we should surface the exception in order to go in to error state.

dstathis (changes requested):

src/agent_workload.py
sed-i: This is [copied](https://github.com/canonical/grafana-agent-k8s-operator/blob/5be2d358b4a2fbb4dca41625626b0a9624ca11d1/src/grafana_agent.py#L418) from gagent.

Seems like:
- It should be WaitingStatus if we're before pebble-ready ("auto resolvable")
- BlockedStatus otherwise

So I agree, all in all this should be BlockedStatus.

src/agent_workload.py
sed-i: TODO:
- log.info the workload version
- do not set the workload here; let charm code access `.version` prop directly
- in this case, charm code should set `n/a` ~(or `multi`/`various`?)~

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+        self._container.restart(self._service_name)
+
+    def _update_config(self) -> None:
+        if not self.is_ready:
+            # Workload is not yet available so no need to update config
+            self.status.config = WaitingStatus("Workload is not yet available")
+            return
+
+        config = self._render_config()  # TODO: Must not be None
+        assert config is not None
```
Abuelodelanada, line 163: Are we capturing possible `AssertionError` exception? :thinking:

src/charm.py
```diff
@@ -66,8 +70,20 @@ def __init__(self, *args):
...
+            container_name="agent",
+            config_getter=lambda: Config(
+                topology=JujuTopology.from_charm(self),
+                scrape_configs=None,  # FIXME generate from memberlist
+                remote_write=self.remote_write_consumer.endpoints,
+                loki_endpoints=self.loki_consumer.loki_endpoints,
+                insecure_skip_verify=True,
+                http_listen_port=3500,
+                grpc_listen_port=3600,
+            ).build(),  # TODO figure out what to do about potential code ordering problem
```
Abuelodelanada, line 84: I would build `Config()` object before, so this section is easy to understand

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Status:
+    """Helping with centralized status setting."""
+
+    def __init__(self, callback: Callable[[StatusBase], None] = lambda _: None):
+        self._config: StatusBase = UnknownStatus()
+        self._callback = callback
```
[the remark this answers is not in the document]
PietroPasotti: I'm thinking we should wait on compound status to land in ops, inventing our own here seems like coinventing the wheel

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+    @property
+    def config(self):
+        """Status for the grafana agent config file."""
+        return self._config
+
+    @config.setter
+    def config(self, value: StatusBase):
+        self._config = value
+        # When status is updated, it is likely desirable to have some kind of side effect.
+        self._side_effect()
```
PietroPasotti, line 38: [nit]: `_side_effect --> _update_status`

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _side_effect(self):
+        logger.debug("Status updated to: %s", self._combined())
+        self._callback(self._combined())
+
+    def _combined(self) -> StatusBase:
```
PietroPasotti, line 44: [nit]: either `_combine()` or `@property\ndef _combined(self)...`

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+        return self._container.can_connect()
+
+    @property
+    def version(self) -> Optional[str]:
+        """Returns the version of the agent.
+
+        Returns:
+            A string equal to the agent version
+        """
+        if not self.is_ready:
```
PietroPasotti, line 126: `is_ready` is not a property (but imho it should be) so this will always be trivially true.

src/agent_workload.py
```diff
@@ -0,0 +1,185 @@
...
+
+        if config == old_config:
+            # Nothing changed, possibly new installation. Move on.
+            self.status.config = ActiveStatus()
+            return
+
+        try:
+            self.write_file(self.CONFIG_PATH, yaml.dump(config))
+        except APIError as e:
+            logger.warning(str(e))
```
PietroPasotti, line 181: if this fails, how does the charm know? and how does it retry?

PietroPasotti (changes requested):
RE the different Config approach, I don't think it's such a different approach. The implementation is different, sure, but the idea is the same. Also AM's config has somewhat different design goals:
- storedstate-backed hash instead of recalculation from filesystem (we could factor that out, of course).
- the config building is somewhat more complicated, not just one big dict.

We should see if we can generalize the two?

sed-i:
Outdated. Closing.

PR #12, opened 2023-11-10
adopt the one-relation model

### LINKED ISSUE #13: rework the charm to the new design
### Enhancement Proposal

The coordinator charm has to:

1. instead of supporting one relation endpoint per role, we should relate the coordinator to the worker over the single interface/endpoint `mimir-cluster`;
1. gather all the roles from relation data and check if the minimal subset conditions are met; if not, set `BlockedStatus` and let the user know with a status message;
1. build the Mimir configuration and update it with the *hash ring (memeberlist)*, composed of all the URLs of the workers (through the [top-level config key `memberlist`](https://grafana.com/docs/mimir/latest/references/configuration-parameters/#memberlist), under `join_members`);
1. send the Mimir configuration to all the workers over relation data

Completing this issue should be enough for a functioning testable charm to be related to a worker.

lucabello:
Closes #13.

Tandem worker PR: https://github.com/canonical/mimir-worker-k8s-operator/pull/13


metadata.yaml
```diff
@@ -33,42 +33,8 @@ resources:
...
-    interface: mimir_worker
-    limit: 1
-  overrides-exporter:
-    interface: mimir_worker
-    limit: 1
-  query-scheduler:
-    interface: mimir_worker
-    limit: 1
+  mimir-cluster:
+    interface: mimir_cluster
```
sed-i, line 37: Add a description field?

src/mimir_coordinator.py
```diff
@@ -49,8 +49,9 @@
 deployment to be considered robust according to the official recommendations/guidelines."""
 
 
-def _endpoint_to_role(endpoint: str) -> MimirRole:
-    return MimirRole(endpoint.replace("-", "_"))
+def _relation_to_role(relation: Relation) -> MimirRole:
+    # TODO: extract the role from the relation
+    return MimirRole(relation.role) # FIXME
```
sed-i, line 54: Should we get this "for free" from the pydantic deserialization?

lucabello: Actually yes; this PR should wait for the interface to be formalized so we can import it and use it here

metadata.yaml
lucabello: Good point, done!

lucabello:
Turning this into a draft PR until we have an interface to import in the charm.

PietroPasotti:
(integration tests are an expected failure ATM)

PietroPasotti (approved):
LGTMimir

PR #17, opened 2023-12-01
add nginx workload

### LINKED ISSUE #14: Add the nginx workload
### Enhancement Proposal

The coordinator should run **nginx** as its workload (with a sidecar grafana agent).

Nginx is a **convenience** feature, so that a user can access all the workers from a single endpoint. The configuration should be taken/adapted from the *distributed Mimir Helm chart*.

lucabello:
## Issue

Closes #14.

## Solution

This PR adds the Nginx workload to the Mimir Coordinator. Some points worth noting:
* the Nginx configuration was taken from the [Mimir HA helm chart](https://github.com/grafana/mimir/blob/63f2f62bf652355b93f4d19937321ab0aafe8c1b/operations/helm/charts/mimir-distributed/values.yaml#L2313) and it's built with [nginxinc/crossplane](https://github.com/nginxinc/crossplane), a python package (provided by nginx) which handles the conversion "python dictionary" => "nginx config";
* I added load balancing to the Nginx config, so that asking something to a **distributor** can reach all the workers that have the **distributor** role; at the moments it's round-robin (deafult), but we can decide what to do with it
* some sections (i.e., `log_verbose`, `resolver`, `basic_auth`) are generated by functions and are supposed to be customizable (e.g., through Juju config options); at the moment those choices are hardcoded (i.e., calling `log_verbose(False)`), however it's easy to modify them in the future to expose those config options if we want to;
* the config currently hardcodes the `http` scheme; this (plus certificate configurations) shall be updated when we implement TLS in the coordinator

### Open questions / future things (FIXME, TODO)

* we could allow the user to specify a custom block for the HTTP Nginx configuration through Juju config (it's something that's present in the Helm chart config, not sure if we need it or not)
* the `noAuthTenant` value (currently set to `FIXMEnoAuthTenant`) should be taken from the Mimir configuration and passed to the Nginx object

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -19,6 +19,7 @@ from charms.prometheus_k8s.v0.prometheus_remote_write import (
     PrometheusRemoteWriteConsumer,
 )
 from mimir_coordinator import MimirCoordinator
+from nginx import Nginx
 from ops.charm import CharmBase, CollectStatusEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, Relation
@@ -32,6 +33,10 @@ class MimirCoordinatorK8SOperatorCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
+
+        self._nginx_container = self.unit.get_container("nginx")
+        self._nginx_config_path = "/etc/nginx/nginx.conf"
+
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.collect_unit_status, self._on_collect_status)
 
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -32,6 +33,10 @@ class MimirCoordinatorK8SOperatorCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
+
+        self._nginx_container = self.unit.get_container("nginx")
+        self._nginx_config_path = "/etc/nginx/nginx.conf"
```
Abuelodelanada, line 38: Do we need this property since we have it in nginx.py?

src/nginx.py
```diff
@@ -0,0 +1,305 @@
...
+                    {"directive": "auth_basic", "args": ['"Mimir"']},
+                    {
+                        "directive": "auth_basic_user_file",
+                        "args": ["/etc/nginx/secrets/.htpasswd"],
+                    },
+                ]
+            return []
+
+        # build the complete configuration
+        full_config = [
```
Abuelodelanada, line 79: Since this `full_config` is a very large config, shouldn't we have it in a separate file just for that?

lucabello: `src/nginx.py` **is** a separate file; all it does is building the Nginx configuration, and it needs to access the `cluster_provider` object (to get the workers' addresses), meaning we'd have to pass it yet another time, and `src/nginx.py` would be almost empty

I'd leave it here, as it's the main point of that extra file! :)

src/charm.py
lucabello: Good eye, I forgot that there; I'll remove it!

Abuelodelanada (approved):

PR #18, opened 2023-12-07
fix nginx config when one or more roles are missing

lucabello:
Currently, when the coordinator is not related to any worker, the **nginx** workload fails to start. This happens because the load balancing section is missing from the configuration, as it's a result of iterating over all the workers.

This PR makes the re-routing `location` directives conditional on the roles being present, so that **nginx** can start even when the coordinator is initially deployed without workers.


lucabello:
@Abuelodelanada @IbraAoad this should fix the issue you guys encountered, feel free to have a look at the PR

Abuelodelanada (approved):

Issue #15, opened 2023-11-13
add TLS support


lucabello:
### Enhancement Proposal

The charm should support TLS through the `certificates` relation, like the other charms in the COS Lite bundle.

It should use the certificate in the **nginx** workload, so that the workers can be reached over TLS.

Abuelodelanada:
Certificates in the *nginx* workload is one aspect of the TLS story of the coordinator.

The other side of this story is generation of [TLS config under `server` section](https://grafana.com/docs/mimir/latest/references/configuration-parameters/#server).This configuration should be generated in the coordinator charm and sent over relation data to workers.

PR #20, opened 2023-12-08
First TLS instrumentation 

Abuelodelanada:
This PR fixes: #15 

- Adds TLS config for the workers
- Adds TLS config for the NGINX workload in the coordinator.

And goes in tandem with:

- https://github.com/canonical/observability-libs/pull/72
- https://github.com/canonical/mimir-worker-k8s-operator/pull/21

## How to test it:

1. Deploy this bundle (You need to pack the coordinator and the worker)
```yaml
bundle: kubernetes
applications:
  ca:
    charm: self-signed-certificates
    channel: edge
    revision: 52
    scale: 1
    constraints: arch=amd64
  coord:
    charm: local:mimir-coordinator-k8s-21
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
    trust: true
  worker:
    charm: local:mimir-worker-k8s-8
    scale: 1
    options:
      all: true
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
    trust: true
relations:
- - coord:mimir-cluster
  - worker:mimir-cluster
- - coord:certificates
  - ca:certificates
```
2. Check that the worker config file has the TLS settings in place: `juju ssh --container mimir worker/0 cat /etc/mimir/mimir-config.yaml`
3. Check the nginx config in the coordinator has SSL: `juju ssh --container nginx coord/0 cat /etc/nginx/nginx.conf`


src/charm.py
```diff
@@ -98,9 +128,15 @@ def _on_config_changed(self, event):
         """Handle changed configuration."""
         self.publish_config()
 
-    def publish_config(self):
+    def _on_server_cert_changed(self, _):
+        self._update_cert()
+        self._on_nginx_pebble_ready(_)
+        self.publish_config(tls=self._is_cert_available)
```
sed-i, line 134: MimirCoordinator take a server cert in its constructor so do we need here something like
```python
self.coordinator.update_server_cert(...)
```

src/charm.py
```diff
@@ -98,9 +128,15 @@ def _on_config_changed(self, event):
...
+
+    def publish_config(self, tls: bool = False):
         """Generate config file and publish to all workers."""
-        mimir_config = self.coordinator.build_config(dict(self.config))
+        mimir_config = self.coordinator.build_config(dict(self.config), tls=tls)
+        logger.warning(mimir_config)
```
sed-i, line 139:
```suggestion
        logger.warning(mimir_config)  # TODO: remove
```

src/charm.py
```diff
@@ -133,12 +169,62 @@ def _on_loki_relation_changed(self, _):
...
+                self.server_cert.key,  # pyright: ignore
+                make_dirs=True,
+            )
+            # Save the CA among the trusted CAs and trust it
+            self._nginx_container.push(
+                ca_cert_path,
+                self.server_cert.ca,  # pyright: ignore
+                make_dirs=True,
+            )
+            # FIXME with the update-ca-certificates machinery prometheus shouldn't need
```
sed-i, line 203: Is this FIXME relevant here?

src/charm.py
```diff
@@ -133,12 +169,62 @@ def _on_loki_relation_changed(self, _):
...
+            #  CA_CERT_PATH.
+            self._nginx_container.push(
+                CA_CERT_PATH,
+                self.server_cert.ca,  # pyright: ignore
+                make_dirs=True,
+            )
+
+            # Repeat for the charm container. We need it there for prometheus client requests.
+            ca_cert_path.parent.mkdir(exist_ok=True, parents=True)
+            ca_cert_path.write_text(self.server_cert.ca)  # pyright: ignore
```
sed-i, line 213: Do we need this for the coordinator?

src/mimir_coordinator.py
```diff
@@ -120,10 +120,20 @@ def build_config(self, _charm_config: Dict[str, Any]) -> Dict[str, Any]:
...
-        if self._tls_requirer:
-            mimir_config["tls_enabled"] = True
-            mimir_config["tls_cert_path"] = self._tls_requirer.cacert
-            mimir_config["tls_key_path"] = self._tls_requirer.key
-            mimir_config["tls_ca_path"] = self._tls_requirer.capath
+        if tls:
+            mimir_config["server"] = {
+                "http_tls_config": {
+                    "cert": self._tls_requirer.cert,
+                    "key": self._tls_requirer.key,
```
sed-i, line 127: - Does this mean the private key will travel over relation data?
- With nginx in place, are we going for e2e tls or just termination?

tests/integration/test_tls.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def get_nginx_config(ops_test: OpsTest):
+    return get_workload_file(
+        ops_test.model_name, mimir_app_name, 0, "nginx", "/etc/nginx/nginx.conf"
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_nginx_config_has_ssl(ops_test: OpsTest):
```
sed-i, line 32: Should we inline and deploy the same bundle you proposed in the PR description?

tox.ini
```diff
@@ -49,7 +49,9 @@ commands =
               --skip {toxinidir}/.mypy_cache \
               --skip {toxinidir}/.ruff_cache \
               --skip {toxinidir}/icon.svg \
-              --skip {toxinidir}/src/grafana_dashboards
+              --skip {toxinidir}/src/grafana_dashboards \
+              --skip {toxinidir}/tox.ini \
+              -L aNULL,anull
```
sed-i, line 54: What's that?

tox.ini
```diff
@@ -69,6 +71,8 @@ deps =
     pytest
     coverage[toml]
     -r{toxinidir}/requirements.txt
+    cryptography
```
sed-i, line 74:
```suggestion
    
    # Binary deps from from charmcraft.yaml
    cryptography
```

src/mimir_coordinator.py
Abuelodelanada: The idea is that we create the config in the coordinator and distribute it to the workers. That's why we are setting everything here.

Now that you mention this, I think that if the communication channel where relation data travels is encrypted, then there is no risk. WDYT?

If we cannot ensure this the process of generating the configuration will be a bit more complex.

As far as I understand for nginx we need e2e, anyway @PietroPasotti or @lucabello may have better insights

tests/integration/test_tls.py
Abuelodelanada: I would add that itest once we have a minimal version of the worker on charmhub so we can avoid the build process of that charm.

tox.ini
Abuelodelanada: This prevents linting error due to:
https://github.com/canonical/mimir-coordinator-k8s-operator/pull/20/files#diff-eab582b474ff3c4ac3d1aba17cc59c806e8dc266e6c48a0ed35d3a60edbcc9d5R341

src/charm.py
```diff
@@ -98,9 +128,14 @@ def _on_config_changed(self, event):
         """Handle changed configuration."""
         self.publish_config()
 
-    def publish_config(self):
+    def _on_server_cert_changed(self, _):
+        self._update_cert()
+        self._on_nginx_pebble_ready(_)
```
dstathis, line 133: I don't like calling a hook callback here directly. Maybe make a function called self.configure() or something and call that in both hooks.

src/mimir_coordinator.py
sed-i: > if the communication channel where relation data travels is encrypted

That is one of the reasons we're moving away from peer data for storing secrets.

src/charm.py
Abuelodelanada: For now I would keep this here, and see how it evolves.
I like the idea of moving this to `MimirCoordinator`, but on the other hand I like to have everything related to certs in the "same" place.

lucabello (approved):

PR #21, opened 2023-12-13
Add S3 relation to mimir coordinator

IbraAoad:
**Description**

Add the ability to integrate the Mimir cluster with external S3 Storage

**Rationale**
Mimir is a highly scalable, highly available solution. As such, using host path- or PVC-backed storage won’t be enough to fulfill the resilience and availability requirements when deploying a cluster with multiple units or applications.


src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
 
     def publish_config(self):
         """Generate config file and publish to all workers."""
-        mimir_config = self.coordinator.build_config(dict(self.config))
+        mimir_config = self.coordinator.build_config(self._s3_storage)
         self.cluster_provider.publish_configs(mimir_config)
 
+    def create_minio_buckets(self, conn: dict, bucket_names: list):
```
sed-i, line 116: Should we add more to the docstring here?

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+                else:
+                    logger.info("Bucket already exists")
+        except S3Error:
+            logger.error("Error creating S3 buckets")
+
+    def _parse_s3_data(self, s3_data):
+        data_passed_dict = dict(item.split(": ") for item in s3_data.split("\n") if item)
+        data_passed_dict["secure"] = data_passed_dict.get("secure", "").lower() == "true"
+
+        return {
```
sed-i, line 138: Where is this schema coming from? Should we have a dataclass or pydantic model for it?

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+            logger.warning("Incoherent deployment: Some required Mimir roles are missing.")
+            return
+        self._process_s3_relation()
+        if not self._s3_storage_data and self.coordinator.is_scaled():
+            logger.warning("Filesystem storage cannot be used with replicated mimir workers")
+            return
+        self.publish_config()
+
+    def _process_s3_relation(self):
+        s3_relation = self.model.get_relation("s3")
```
sed-i, line 158: This will fail if there are multiple s3 relations.
Need to either add `limit:` to metadata or iterate over all relations.

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+        if s3_relation and s3_relation.app:
+            data_passed = s3_relation.data.get(s3_relation.app, {}).get("data")
+            if data_passed:
```
sed-i, line 161: Could simplify with a walrus, 

```suggestion
            if data_passed := s3_relation.data.get(s3_relation.app, {}).get("data"):
```

but we should probably instantiate a pydantic model from the relation data, instead of fetching `.get("data")` ourselves here.

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+        if not self.coordinator.is_coherent():
+            logger.warning("Incoherent deployment: Some required Mimir roles are missing.")
+            return
+        self._process_s3_relation()
+        if not self._s3_storage_data and self.coordinator.is_scaled():
+            logger.warning("Filesystem storage cannot be used with replicated mimir workers")
+            return
+        self.publish_config()
+
+    def _process_s3_relation(self):
```
sed-i, line 157: I wonder if we need this method. Seems like it's just copying data from relation data to a class property.
Haven't checked but perhaps something like the following makes sense?
```python
# on relation changed:
try:
    data = SomePydanticBaseModel(s3_relation.data.get(s3_relation.app, {}))
except SomePydanticError:
    self.cleanup_previous_side_effect()
else:
    self.some_side_effect()
```

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+                s3_parsed_data = self._parse_s3_data(data_passed)
+                self.create_minio_buckets(s3_parsed_data, ["mimir"])
+                self._s3_storage_data = s3_parsed_data
+                return
+        self._s3_storage_data = None
+
+    def _on_s3_created(self, _):
+        s3_relation = self.model.get_relation("s3")
+        if s3_relation:
+            s3_relation.data[self.model.app]["_supported_versions"] = "- v1"
```
sed-i, line 171: Should we manage this with a pydantic model?

src/charm.py
```diff
@@ -100,23 +110,86 @@ def _on_config_changed(self, event):
...
+
+    def _on_s3_broken(self, _):
+        if not self.coordinator.is_coherent():
+            logger.warning("Incoherent deployment: Some required Mimir roles are missing.")
```
sed-i, line 175: From past experience it's best when status is set in a centralized way.
I wonder if [collect_status](https://github.com/canonical/operator/pull/954) would work here.

Ah I see you use it. The thing with collect_status is that it's an "all in" solution. These statuses here would have no effect now.

src/mimir_coordinator.py
```diff
@@ -69,6 +69,10 @@ def is_coherent(self) -> bool:
         roles: Iterable[MimirRole] = self._cluster_provider.gather_roles().keys()
         return set(roles).issuperset(MINIMAL_DEPLOYMENT)
 
+    def is_scaled(self) -> bool:
+        """Return True if more than 1 worker are forming the mimir cluster."""
+        return len(list(self._cluster_provider.gather_addresses())) > 1
```
sed-i, line 74: Why do we use "gather addresses" (indirect method) instead of looking at the number of units in the relation (direct method) to determine if scaled?

requirements.txt
```diff
@@ -2,3 +2,5 @@ ops
 pydantic
 # crossplane is a package from nginxinc to interact with the Nginx config
 crossplane
+
+minio
```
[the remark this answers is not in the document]
dstathis: Agreed. Would this work with other S3 providers?

metadata.yaml
```diff
@@ -29,7 +29,7 @@ resources:
 
 requires:
   s3:
-    interface: s3
+    interface: object-storage
```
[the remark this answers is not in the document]
IbraAoad: Switched to [s3 integrator](https://charmhub.io/s3-integrator)

requirements.txt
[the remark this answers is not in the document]
IbraAoad: Removed, but on a side note mimir expects an s3 bucket to exist and I'm wondering which of the below make sense

1. Leave it as the responsibility of the user to actually create a bucket before hand and also to ensure that the s3 connection provided is valid
2. Make use of something like [boto3](https://pypi.org/project/boto3/) to validate the s3 connection and create the bucket if it doesn't exists, but this will make boto3 a dependency tho

src/charm.py
```diff
@@ -52,11 +54,27 @@ def __init__(self, *args):
             self._on_nginx_pebble_ready,
         )
 
+        self._s3_storage_data = None
```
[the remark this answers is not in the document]
IbraAoad: switched to an empty pydantic model at initialization

src/charm.py
```diff
@@ -52,11 +54,27 @@ def __init__(self, *args):
...
+        )
+
+        self.framework.observe(
+            self.on.s3_relation_changed,  # pyright: ignore
+            self._on_mimir_cluster_changed,
+        )
+
+        self.framework.observe(
+            self.on.s3_relation_broken,  # pyright: ignore
+            self._on_s3_broken,
```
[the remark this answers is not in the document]
IbraAoad: Switched to the [credentials_gone](https://github.com/canonical/s3-integrator/blob/main/lib/charms/data_platform_libs/v0/s3.py#L559) provided by the s3 interface

src/charm.py
IbraAoad: Multiline text coming through from [Charmed Minio](https://github.com/canonical/minio-operator/blob/main/src/charm.py#L161-L172), removed now after moving to s3-intgerator

src/charm.py
IbraAoad: Moved the check inside the collect_status

src/mimir_coordinator.py
IbraAoad: Problem here is that only the coordinator (A workload-less charm) will connect to s3, generate the config and then pass it to the workers over the coordinator-worker relation, only the workers will be replicated and in some cases there might be different apps with several units all forming the **mimir-worker-cluster**, so this was the easiest way tbh to find if the workers are scaled. I'm open to better suggestions tho :D

Abuelodelanada: Well, that's not strictly true. The coordinator will run nginx.

src/mimir_coordinator.py
```diff
@@ -81,49 +86,131 @@ def is_recommended(self) -> bool:
...
-                "data_dir": str(self._root_data_dir / "data-alertmanager"),
-            },
-            "compactor": {
-                "data_dir": str(self._root_data_dir / "data-compactor"),
+            "alertmanager": self._build_alertmanager_config(),
+            "alertmanager_storage": self._build_alertmanager_storage_config(),
+            "compactor": self._build_compactor_config(),
+            "ruler": self._build_ruler_config(),
+            "ruler_storage": self._build_ruler_storage_config(),
+            "blocks_storage": self._build_blocks_storage_config(),
```
Abuelodelanada, line 101: I like this refactor.
But if these methods do not receive arguments, shouldn't be properties?

IbraAoad: Thought of that but decided not to switch them cause I think we might be adding to the configuration of each component with future requirements, for ex [alertmanager](https://grafana.com/docs/mimir/latest/references/configuration-parameters/#alertmanager) has tons of configuration that we haven't used yet and that might require future args passed to the functions, what I'm still not sure of tho is whether the config would reach the workers with no changes required to the config fields or if there will still be some changes the worker charm will have to do like the root directory for instance in the current case

src/charm.py
```diff
@@ -52,11 +56,17 @@ def __init__(self, *args):
             self._on_nginx_pebble_ready,
         )
 
+        self._s3_storage_data = _S3StorageBackend()
```
PietroPasotti, line 59: the identifier name should match the name of the type:
either `_s3_storage_backend = ...` or `... = _S3StorageData`

requirements.txt
```diff
@@ -1,4 +1,4 @@
 ops
 pydantic
 # crossplane is a package from nginxinc to interact with the Nginx config
```
PietroPasotti, line 3: more useful to add a comment about what module needs it

metadata.yaml
```diff
@@ -32,8 +32,9 @@ requires:
     interface: s3
     limit: 1
     description: |
-      The coordinator obtains storage info on behalf of the workers, and
-      forwards all workers the storage details over mimir-worker.
+      The coordinator obtains storage information on behalf of the workers and forwards all workers
```
PietroPasotti, line 35: this feels like a good comment to understand the design/implementation of Mimir, but not a good public description.
I'd replace it with something like:
`interface to grant Mimir access to a s3 bucket for data storage.`

limit: 1 is self-explanatory IMHO.

src/charm.py
```diff
@@ -52,11 +56,17 @@ def __init__(self, *args):
...
         self.framework.observe(
             self.on.mimir_cluster_relation_changed,  # pyright: ignore
             self._on_mimir_cluster_changed,
         )
 
+        self.framework.observe(
+            self.s3_requirer.on.credentials_changed, self._on_mimir_cluster_changed
```
PietroPasotti, line 66: - add a `self._on_s3_requirer_credentials_changed` observer, and from that observer, invoke whatever you need to invoke (and observers shouldn't call one another). see the "Only event handlers should be event-aware" section of https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619

src/charm.py
```diff
@@ -100,23 +98,47 @@ def _on_config_changed(self, event):
 
     def publish_config(self):
         """Generate config file and publish to all workers."""
-        mimir_config = self.coordinator.build_config(dict(self.config))
+        mimir_config = self.coordinator.build_config(self._s3_storage_data)
```
PietroPasotti, line 101: this is a race condition waiting to happen. Instead of 'persisting' the data in an instance attr, why don't we make `_s3_connection_info` **return** `S3StorageData` and pass it directly into build_config here?

src/charm.py
```diff
@@ -100,23 +98,47 @@ def _on_config_changed(self, event):
...
         self.cluster_provider.publish_configs(mimir_config)
 
+    def _s3_connection_info(self):
+        """Parse a _S3StorageBackend object from relation data."""
+        if not self.s3_requirer.relations:
+            self._s3_storage_data = _S3StorageBackend()
+            return
+        raw = self.s3_requirer.get_s3_connection_info()
+        raw["access_key"], raw["secret_key"] = raw.pop("access-key", ""), raw.pop("secret-key", "")
+        self._s3_storage_data = _S3StorageBackend(**raw)
```
PietroPasotti, line 111: Since we use Pydantic, you can use aliases to spare you the need of doing this translation shenanigans

src/charm.py
```diff
@@ -100,23 +98,47 @@ def _on_config_changed(self, event):
...
     def _on_mimir_cluster_changed(self, _):
-        if self.coordinator.is_coherent():
-            logger.info("mimir deployment coherent: publishing configs")
-            self.publish_config()
-        else:
-            logger.warning("this mimir deployment is incoherent")
+        if not self.coordinator.is_coherent():
+            logger.warning("Incoherent deployment: Some required Mimir roles are missing.")
+            return
+        self._s3_connection_info()
```
PietroPasotti, line 117: what is this call doing? are you implicitly relying on a side-effect?

src/charm.py
```diff
@@ -100,23 +98,47 @@ def _on_config_changed(self, event):
...
+        if self._s3_storage_data == _S3StorageBackend() and self.coordinator.is_scaled():
```
PietroPasotti, line 118: here I'd do something like add a property `self._is_s3_available` to check that we have a s3 relation and we have a bucket configured, instead of relying on that instance attr

src/charm.py
```diff
@@ -100,23 +98,47 @@ def _on_config_changed(self, event):
...
         if not self.coordinator.is_coherent():
             event.add_status(
-                BlockedStatus(
-                    "Incoherent deployment: you are " "lacking some required Mimir roles"
-                )
+                BlockedStatus("Incoherent deployment: you are lacking some required Mimir roles")
+            )
+        if self._s3_storage_data == _S3StorageBackend() and self.coordinator.is_scaled():
+            event.add_status(
+                BlockedStatus("Missing s3 relation, replicated units must use S3 storage.")
```
PietroPasotti, line 141:
```suggestion
                BlockedStatus("When multiple units of Mimir are deployed, you must add a S3 relation. S3 relation missing.")
```

src/mimir_config.py
```diff
@@ -84,11 +87,11 @@ class Alertmanager(BaseModel):
 
 
 class _S3StorageBackend(BaseModel):
```
PietroPasotti, line 89: If we're importing this from another module, it should not start with an underscore.

[also, a nitpick] I'd rename it to `S3BucketConfig` or `S3StorageBackendData`, because it's not really a 'backend': it's the config/data you need to connect to one

src/mimir_config.py
```diff
@@ -84,11 +87,11 @@ class Alertmanager(BaseModel):
...
-    endpoint: str
-    access_key_id: str
-    secret_access_key: str
-    insecure: bool = False
-    signature_version: str = "v4"
+    access_key: str = ""
```
PietroPasotti, line 90: if these are mandatory fields, as I expect, they shouldn't have default values. Am I missing something?

src/mimir_coordinator.py
```diff
@@ -69,6 +70,10 @@ def is_coherent(self) -> bool:
         roles: Iterable[MimirRole] = self._cluster_provider.gather_roles().keys()
         return set(roles).issuperset(MINIMAL_DEPLOYMENT)
 
+    def is_scaled(self) -> bool:
```
PietroPasotti, line 73: since we're checking whether the workers are scaled, I'd do instead:
`number_of_workers(self) -> int` so you can write more explicit/transparent code to use it

src/mimir_coordinator.py
```diff
@@ -81,49 +86,130 @@ def is_recommended(self) -> bool:
                 return False
         return True
 
-    def build_config(self, _charm_config: Dict[str, Any]) -> Dict[str, Any]:
+    def build_config(self, s3_data: _S3StorageBackend) -> Dict[str, Any]:
```
PietroPasotti, line 89: again try to use variable names matching the type names:
if it's a 'backend', why are we calling it a 'data'?

either `s3_data: S3Data` or `s3_backend: S3Backend` (I suggest the former)

src/mimir_coordinator.py
```diff
@@ -81,49 +86,130 @@ def is_recommended(self) -> bool:
...
+                "endpoint": s3_data.endpoint,
+                "access_key_id": s3_data.access_key,
+                "secret_access_key": s3_data.secret_key,
+                "bucket_name": s3_data.bucket,
+                "region": s3_data.region,
+            },
+        }
 
-        return mimir_config
+    def _update_s3_storage_config(
```
PietroPasotti, line 199: not sure what this method is doing and why

PietroPasotti (changes requested):
Great to see the ball rolling but I see a few issues with the implementation.
Some stylistic feedback, and some charming best practices we should enforce.

Great job overall! Looking forward to the next steps :)

src/mimir_coordinator.py
lucabello: I think the naming of the method is what's causing the confusion here, so I need a small clarification :)

If the correct semantics are expressed in the docstring (which I believe is the case), and we want to know *whether multiple worker units (on any worker app) are related to the coordinator*, then this makes sense (though it could be changed with `units in relation > 1 OR relations > 1`.

In this context, we'd want the method to return `True` even when two single-unit workers are related to the coordinator.

**scaled** here means (IIUC) having multiple workers; every role could still have one unit (so **not scaled**), but the cluster would still be (according to this method).

Am I getting that right?

IbraAoad: Got the function name changed now to `has_multiple_workers` and I believe I haven't expressed it correctly but mainly even if two single-unit workers are related to the coordinator we would still need mimir to use S3

requirements.txt
IbraAoad: I actually haven't added any new dependency, not sure why it's showing that there's a change in the file

src/mimir_config.py
IbraAoad: The default values are used to facilitate checking if the model instance has been initialized with actual values. This allows us to determine if the S3 configuration is connected or not.

```
    def _get_s3_storage_config(self):
        """Retrieve S3 storage configuration."""
        if not self.s3_requirer.relations:
            return _S3ConfigData()
        raw = self.s3_requirer.get_s3_connection_info()
        raw["access_key"], raw["secret_key"] = raw.pop("access-key", ""), raw.pop("secret-key", "")
        return _S3ConfigData(**raw)
```

the above is triggered whenever a new worker joins the cluster, returning the model with default values helps in telling that Mimir isn't using S3 hence blocking the state

src/mimir_config.py
```diff
@@ -84,11 +87,11 @@ class Alertmanager(BaseModel):
...
+    endpoint: str = ""
+    secret_key: str = ""
+    bucket: str = ""
```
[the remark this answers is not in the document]
IbraAoad: The [S3 requirer ](https://github.com/canonical/s3-integrator/blob/main/lib/charms/data_platform_libs/v0/s3.py#L578), takes in the s3 bucket name as a param, if not passed it sets it to `f"relation-{event.relation.id}"`, the glob is to ensure that the bucket that will be passed in the relation data won't be randomly generated and eventually the bucket var will take it's value from the relation data which will be the glob value, I was thinking of dumping the glob in favor of setting the bucket default value as "mimir" in the model, but went with this to have the bucket name consistent between the relation data and the mimir config

src/mimir_coordinator.py
IbraAoad: changed the name to has_multiple_workers but retained the logic, tbh I'm not sure if returning the actual number of workers would add anymore benefit cause mimir would need s3 if it's more than more worker anyways

src/mimir_coordinator.py
IbraAoad: added some docstrings to explain, this mainly pops any occurrence of filesystem in the mimir config and replaces it with a storage_prefix to be used with the s3 bucket and it only gets triggered if the S3 pydantic model has actual values and not defaults

requirements.txt
lucabello: True, this is a dependency I added, and it's needed to build the Nginx config

src/mimir_config.py
PietroPasotti: but still, are we sure that the s3 databag is allowed to not contain any of these fields? If the databag is empty, it's likely not valid and we should abort.
Otherwise we'll call build_config with empty strings all over and we'll generate an invalid config.
I suggest you use an approach like [this](https://github.com/canonical/tempo-k8s-operator/blob/main/lib/charms/tempo_k8s/v1/tracing.py#L125) to validate the databags. Usually we do this in the charm lib but in this case it's an external library... Perhaps worth adding a separate `s3_interface.py` module to contain this?

src/mimir_coordinator.py
PietroPasotti: still feels ugly to count the remote units by counting the addresses

more work, but also more transparent, would be to do something like

`sum(len(r.units) for r in self.model.relations["mimir_cluster"])`

with the caveat that you probably need to filter out "this app's units"

src/charm.py
```diff
@@ -77,46 +88,64 @@ def __init__(self, *args):
...
         """Handle start event."""
         if not self.coordinator.is_coherent():
             event.add_status(
-                BlockedStatus(
-                    "Incoherent deployment: you are " "lacking some required Mimir roles"
-                )
+                BlockedStatus("Incoherent deployment: you are lacking some required Mimir roles")
+            )
+        s3_config_data = self._get_s3_storage_config()
+        if s3_config_data == _S3ConfigData() and self.coordinator.has_multiple_workers():
```
PietroPasotti, line 146: Rather than doing this, I'd have _get_s3_storage_config raise an exception if it fails to validate the databag. It's pydantic after all. The idea is not to always generate a model no matter what, but to only generate a model if the data is valid!

PietroPasotti (changes requested):
Some issues are not resolved.

src/mimir_config.py
```diff
@@ -83,19 +89,29 @@ class Alertmanager(BaseModel):
...
-    access_key_id: str
-    secret_access_key: str
-    insecure: bool = False
-    signature_version: str = "v4"
+    secret_access_key: str = Field(alias="secret-key")
+    bucket_name: str = Field(alias="bucket")
+    region: str = ""
+
+    @classmethod
+    def load_and_dump_to_dict(cls, raw_s3_config_data: dict) -> Optional[dict]:
```
PietroPasotti, line 101: we don't need this method: pass the `_S3ConfigData` object around as-is

src/mimir_coordinator.py
```diff
@@ -81,49 +81,125 @@ def is_recommended(self) -> bool:
                 return False
         return True
 
-    def build_config(self, _charm_config: Dict[str, Any]) -> Dict[str, Any]:
+    def build_config(self, s3_config_data: Optional[dict]) -> Dict[str, Any]:
```
PietroPasotti, line 84:
```suggestion
    def build_config(self, s3_config_data: Optional[_S3ConfigData]) -> Dict[str, Any]:
```

src/mimir_coordinator.py
```diff
@@ -81,49 +81,125 @@ def is_recommended(self) -> bool:
...
-                },
-            }
-            mimir_config["blocks_storage"] = {
-                "s3": {"bucket_name": s3_config.blocks_bucket_name}  # e.g. 'mimir-blocks'
-            }
-
-        # memberlist config for gossip and hash ring
-        mimir_config["memberlist"] = {
-            "join_members": list(self._cluster_provider.gather_addresses())
+    def _build_s3_storage_config(self, s3_config_data: dict) -> Dict[str, Any]:
```
PietroPasotti, line 180:
```suggestion
    def _build_s3_storage_config(self, s3_config_data: _S3ConfigData) -> Dict[str, Any]:
```

src/mimir_coordinator.py
```diff
@@ -81,49 +81,125 @@ def is_recommended(self) -> bool:
...
+        return {
+            "backend": "s3",
+            "s3": s3_config_data,
```
PietroPasotti, line 183:
```suggestion
            "s3": s3_config_data.model_dump()
```

PietroPasotti (changes requested):
It's better to pass around typed data structures instead of untyped, opaque ones whenever possible.

You get better type hints and type checks if you pass around `_S3ConfigData` rather than its dictified version. I'd advise to postpone dumping it to dict until you really need the raw data structure.

To see how much this simplification helps with typing, add 

`reportUnknownParameterType = true`

to `pyproject.toml[tool.pyright]` and run `tox -e static-charm`

(which we should put there in a different PR)

src/charm.py
```diff
@@ -10,19 +10,24 @@
...
+from charms.data_platform_libs.v0.s3 import (
+    S3Requirer,
+)
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LokiPushApiConsumer
 from charms.mimir_coordinator_k8s.v0.mimir_cluster import MimirClusterProvider
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
     PrometheusRemoteWriteConsumer,
 )
+from mimir_config import BUCKET_NAME, S3_RELATION_NAME, _S3ConfigData
```
PietroPasotti, line 24: I think I said this before, but since it's a 'public' class (we're importing it here), we should remove the underscore prefix

src/mimir_config.py
```diff
@@ -83,19 +89,20 @@ class Alertmanager(BaseModel):
...
+    model_config = {"populate_by_name": True}
+    access_key_id: str = Field(alias="access-key")
     endpoint: str
-    access_key_id: str
-    secret_access_key: str
-    insecure: bool = False
-    signature_version: str = "v4"
+    secret_access_key: str = Field(alias="secret-key")
+    bucket_name: str = Field(alias="bucket")
+    region: str = ""
```
PietroPasotti, line 98: why does region still have an empty string as default?

PietroPasotti (approved):
Nice! We're getting there. 
No need for another pass from my side, but please check that you haven't forgotten that default arg in the _S3ConfigData object :)

PietroPasotti:
(and before we merge, let's make sure the integration tests are green)

src/mimir_config.py
IbraAoad: Not a mandatory field, a user might not define a region and that shouldn't stop the connection

requirements.txt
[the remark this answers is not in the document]
IbraAoad: Will go with option 1 and we can have a separate issue for s3 bucket creation and connection validation if needed

IbraAoad:
Integration tests are failing and are out of this PR scope, merging this PR and Integration will be tackled separately

PR #24, opened 2024-01-05
fixed scenario tests and made pyright checks stricter

PietroPasotti:
found some broken scenarios and added reportUnknownParameterType = true to pyright config


PietroPasotti:
itests are unrelatedly borky, we can merge on red in this one as it's out of scope

IbraAoad (approved):

PR #26, opened 2024-01-08
Hardcoding the root_dir to match the worker charm

IbraAoad:
Hard-coding the common root data directory in Mimir config generated by the coordinator to be identical to the one in the worker's metadata before passing it over relation data


Abuelodelanada (approved):

dstathis (approved):
I think somewhere in /var would be more appropriate but it doesn't really matter so up to you.

PR #28, opened 2024-01-09
Add charm tracing

mmkay:
This PR adds charm tracing support to mimir-coordinator. It doesn't use TLS on the communication with tempo yet - this will be done later in a follow-up PR.

# How to use tracing

In the model where you have your cos stack, run the following:
```
juju deploy tempo-k8s
juju relate tempo-k8s:grafana-source grafana:grafana-source
<deploy mimir>
juju relate mimir-coordinator-k8s:tracing tempo-k8s:tracing
```

Then in Grafana you can browse your charm traces:
Screenshot from 2024-01-09 17-56-04

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -59,7 +59,7 @@ requires:
     interface: tracing
     limit: 1
     description: |
-      Enables sending traces for charm troubleshooting.
+      Enables sending traces to the tracing backend.
 
 provides:
   mimir-cluster:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -55,6 +55,12 @@ requires:
     description: |
       Ingress-per-app, to load-balance across multiple units of the coordinator.
 
+  tracing:
+    interface: tracing
+    limit: 1
+    description: |
+      Enables sending traces for charm troubleshooting.
+
 provides:
   mimir-cluster:
     interface: mimir_cluster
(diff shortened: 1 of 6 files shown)


metadata.yaml
```diff
@@ -55,6 +55,12 @@ requires:
     description: |
       Ingress-per-app, to load-balance across multiple units of the coordinator.
 
+  tracing:
+    interface: tracing
+    limit: 1
+    description: |
+      Enables sending traces for charm troubleshooting.
```
PietroPasotti, line 62: for now this is fine, but keep in mind some day we might use tracing to monitor more than the charm itself (nginx?)

e.g. send the tracing info forward to the workers so we can send mimir traces to tempo, and the mirmir_worker charm as well.

src/charm.py
```diff
@@ -185,6 +199,14 @@ def _on_nginx_pebble_ready(self, _event: ops.PebbleReadyEvent) -> None:
         self._nginx_container.add_layer("nginx", self.nginx.layer, combine=True)
         self._nginx_container.autostart()
 
+    @property
+    def tempo(self) -> Optional[str]:
```
PietroPasotti, line 203:
```suggestion
    def tempo_endpoint(self) -> Optional[str]:
```

PietroPasotti (approved):
some cosmetic thoughts, but looks good overall!
GJ

src/charm.py
mmkay: Done.

metadata.yaml
mmkay: True, and that's absolutely wanted for future. Changed the description to be more universal.

PR #30, opened 2024-01-17
Add support for Pebble Log forwarding in the workers

IbraAoad:
This PR allows the coordinator to get the Loki endpoints from the `loki_push_api` interfaces (and thus from Loki/Grafana Agent relations) and forward them to the worker over the `mimir-cluster` relation data.

The work in this PR should be repeated in the **charm-relation-interfaces**, in preparation for the Mimir HA charms to not use the `mimir-cluster.py` library anymore.


lucabello:
With the new implementation of the `LogForwarder` class, we probably want to change the content of relation data to be closer to what we need there; I'd pass, from the coordinator, a dictionary of {loki_unit_name: url}, so we can pass it directly to the log forwarder

lucabello:
I think that publishing the Loki endpoint in the same function as the Mimir config is creating some problems; I'll change things around a bit :)

src/charm.py
```diff
@@ -218,6 +238,24 @@ def server_cert_path(self) -> Optional[str]:
     # === UTILITY METHODS === #
     ###########################
 
+    def _update_mimir_cluster(self):  # common exit hook
+        # TODO: does this prevent disabling log forwarding on relation-broken ???
```
PietroPasotti, line 242: this is probably something we should check before we merge

src/charm.py
```diff
@@ -218,6 +238,24 @@ def server_cert_path(self) -> Optional[str]:
...
+        # if not self.coordinator.is_coherent():
```
PietroPasotti, line 243: this is also applicable imho. why is it commented out?

tests/scenario/test_mimir_cluster_interface.py
```diff
@@ -114,9 +114,10 @@ def test_address_collection(workers_addresses):
 
 def test_requirer_getters():
     cfg = {"a": "b"}
+    lgs = {}
     relation = Relation(
         "mimir-cluster-require",
-        remote_app_data=MimirClusterProviderAppData(mimir_config=cfg).dump(),
+        remote_app_data=MimirClusterProviderAppData(mimir_config=cfg, loki_endpoints=lgs).dump(),
```
PietroPasotti, line 120: this change should not be necessary, right?

src/charm.py
```diff
@@ -173,6 +167,7 @@ def _on_remote_write_endpoints_changed(self, _):
 
     def _on_loki_relation_changed(self, _):
         # TODO Update rules relation with the new list of Loki push-api endpoints
+        self._update_mimir_cluster()
         pass
```
PietroPasotti, line 171: you can drop the `pass` now

src/charm.py
```diff
@@ -173,6 +167,7 @@ def _on_remote_write_endpoints_changed(self, _):
 
     def _on_loki_relation_changed(self, _):
         # TODO Update rules relation with the new list of Loki push-api endpoints
```
PietroPasotti, line 169: can drop the todo, I guess we're doing it now

src/charm.py
```diff
@@ -218,6 +238,24 @@ def server_cert_path(self) -> Optional[str]:
...
 
+    def _update_mimir_cluster(self):  # common exit hook
+        # TODO: does this prevent disabling log forwarding on relation-broken ???
+        # if not self.coordinator.is_coherent():
+        #     return
+        tls = self._is_cert_available
+
+        s3_config_data = self._get_s3_storage_config()
+
+        self.cluster_provider.publish_data(
```
PietroPasotti, line 249: I'd add a couple of comments here or a docstring to explain the holistic logic going on here. Be kind to your future self

src/charm.py
PietroPasotti: note the previous implementation had this guard and:
```
logger.warning("Incoherent deployment: Some required Mimir roles are missing.")
```

I vote we keep it

PietroPasotti (changes requested):
looking good! couple of issues.

tests/scenario/test_mimir_cluster_interface.py
lucabello: Right, this is a super old change I didn't even notice was there :)

src/charm.py
lucabello: Right, so this is not a problem anymore because the worker will (by itself) clear the log forwarding endpoints on relation-broken, instead of relying on the coordinator to empty the databag.

src/charm.py
lucabello: Answered in another conversation, but this doesn't happen anymore since the worker autonomously disables log forwarding on `mimir-cluster-relation-broken`

Abuelodelanada (approved):

PR #38, opened 2024-03-01
remove lib and move functionality to src/mimir_coordinator.py

### LINKED ISSUE #19: Stop using the mimir_cluster library.
### Enhancement Proposal

The interface definition is moving in to [charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces) so there is really no reason to have a library.

Depends on canonical/charm-relation-interfaces#129

Abuelodelanada:
This PR fixes #19


src/mimir_cluster.py
```diff
@@ -0,0 +1,283 @@
...
+    for role in roles:
+        if role in META_ROLES:
+            expanded_roles.update(META_ROLES[role])
+        else:
+            expanded_roles.add(role)
+    return expanded_roles
+
+
+class MimirClusterProvider(Object):
+    """Mimir Cluster Provider side class."""
```
PietroPasotti, line 78:
```suggestion
    """``mimir-cluster`` provider endpoint wrapper."""
```

src/mimir_cluster.py
```diff
@@ -0,0 +1,283 @@
+#!/usr/bin/env python3
+# Copyright 204 Canonical
+# See LICENSE file for licensing details.
+
+"""Mimir cluster."""
```
PietroPasotti, line 5:
```suggestion
"""This module contains an endpoint wrapper class for the provider side of the ``mimir-cluster`` relation.

As this relation is cluster-internal and not intended for third-party charms to interact with `mimir-coordinator-k8s`, its only user will be the mimir-coordinator-k8s charm. As such, it does not live in a charm lib as most other relation endpoint wrappers do.
"""
```

PietroPasotti (approved):
couple of minor nits (of which one is probably my past self's own fault)
but looks good. Assuming there've been no code changes to the mimir-cluster lib code itself?

Abuelodelanada:
> couple of minor nits (of which one is probably my past self's own fault) but looks good. Assuming there've been no code changes to the mimir-cluster lib code itself?

No, there were no code changes, just moved the code need by the coordinator.

In another PR we have to move the code needed byt the worker to the worker charm.

Besides we need to move the scenario test to the charm-relation-interface, since in that test we are testing the cluster interface.

PR #41, opened 2024-03-07
add replicas config for monolithic deployment

lucabello:
When a role between **ingester**, **alertmanager**, and **store_gateway** has less than 3 worker units, we need to set the `replication_factor` for that role to one.  
If we have 3 or more units, we should stick to the default value of 3 for all of them.

This PR also sets the `cluster_label` parameter to a value extracted from the coordinator's topology.

### DIFF
--- a/src/mimir_coordinator.py
+++ b/src/mimir_coordinator.py
@@ -108,8 +108,10 @@ class MimirCoordinator:
             "alertmanager": self._build_alertmanager_config(),
             "alertmanager_storage": self._build_alertmanager_storage_config(),
             "compactor": self._build_compactor_config(),
+            "ingester": self._build_ingester_config(),
             "ruler": self._build_ruler_config(),
             "ruler_storage": self._build_ruler_storage_config(),
+            "store_gateway": self._build_store_gateway_config(),
             "blocks_storage": self._build_blocks_storage_config(),
             "memberlist": self._build_memberlist_config(),
         }
@@ -141,9 +143,17 @@ class MimirCoordinator:
     # data_dir:
     # The Mimir Alertmanager stores the alerts state on local disk at the location configured using -alertmanager.storage.path.
     # Should be persisted if not replicated
+
+    # sharding_ring.replication_factor:
+    # (advanced) The replication factor to use when sharding the alertmanager.
+
     def _build_alertmanager_config(self) -> Dict[str, Any]:
+        alertmanager_scale = len(
+            self._cluster_provider.gather_addresses_by_role()[MimirRole.alertmanager]
+        )
         return {
             "data_dir": str(self._root_data_dir / "data-alertmanager"),
+            "sharding_ring": {"replication_factor": 1 if alertmanager_scale < 3 else 3},
         }
 
     # filesystem: dir
@@ -163,6 +173,14 @@ class MimirCoordinator:
             "data_dir": str(self._root_data_dir / "data-compactor"),
         }
 
+    # ring.replication_factor: int
+    # Number of ingesters that each time series is replicated to. This option
+    # needs be set on ingesters, distributors, queriers and rulers when running in
+    # microservices mode.
+    def _build_ingester_config(self) -> Dict[str, Any]:
+        ingester_scale = len(self._cluster_provider.gather_addresses_by_role()[MimirRole.ingester])
+        return {"ring": {"replication_factor": 1 if ingester_scale < 3 else 3}}
+
     # rule_path:
     # Directory to store temporary rule files loaded by the Prometheus rule managers.
     # This directory is not required to be persisted between restarts.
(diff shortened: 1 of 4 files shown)


src/mimir_coordinator.py
```diff
@@ -163,6 +173,14 @@ def _build_compactor_config(self) -> Dict[str, Any]:
             "data_dir": str(self._root_data_dir / "data-compactor"),
         }
 
+    # ring.replication_factor: int
+    # Number of ingesters that each time series is replicated to. This option
+    # needs be set on ingesters, distributors, queriers and rulers when running in
+    # microservices mode.
+    def _build_ingester_config(self) -> Dict[str, Any]:
+        ingester_scale = len(self._cluster_provider.gather_addresses_by_role()[MimirRole.ingester])
+        return {"ring": {"replication_factor": 1 if ingester_scale < 3 else 3}}
```
mmkay, line 182: Could we use a constant instead of the magic numbers here, in line 156 and 200?

tests/unit/test_config.py
```diff
@@ -105,7 +108,10 @@ def test_ne_update_s3_storage_config(self):
     def test_build_memberlist_config(self):
         self.cluster_provider.gather_addresses.return_value = ["address1", "address2"]
         memberlist_config = self.coordinator._build_memberlist_config()
-        expected_config = {"join_members": ["address1", "address2"]}
+        expected_config = {"cluster_label": "something", "join_members": ["address1", "address2"]}
+        self.assertIn("cluster_label", expected_config)
+        memberlist_config.pop("cluster_label")
+        expected_config.pop("cluster_label")
```
IbraAoad, line 114: Can we assert some dummy labels instead?

lucabello: Yes, and I'm changing that; however the function is still gonna look a bit awkward; this is because we are patching out the entire `_cluster_provider`, and the `cluster_label` comes from some properties nested deep inside that object (`.cluster_provider._charm.model.name`, etc...)

src/mimir_coordinator.py
lucabello: Yes and no, as the `1` will probably stay the same, but I'm changing the expression to:
```python
{"replication_factor": 1 if ingester_scale < REPLICATION_MIN_WORKERS else DEFAULT_REPLICATION}
```

I've also documented those two variables so it's much clearer what they do

src/mimir_coordinator.py
```diff
@@ -141,9 +149,20 @@ def _build_tls_config(self) -> Dict[str, Any]:
     # data_dir:
     # The Mimir Alertmanager stores the alerts state on local disk at the location configured using -alertmanager.storage.path.
     # Should be persisted if not replicated
+
+    # sharding_ring.replication_factor: int
+    # (advanced) The replication factor to use when sharding the alertmanager.
     def _build_alertmanager_config(self) -> Dict[str, Any]:
+        alertmanager_scale = len(
+            self._cluster_provider.gather_addresses_by_role()[MimirRole.alertmanager]
+        )
```
Abuelodelanada, line 158: As far as I understand, if `gather_addresses_by_role()` may return an empty dict. If that situation occurs, a `KeyError` exception will be raised.

```suggestion
        alertmanager_scale = len(
            self._cluster_provider.gather_addresses_by_role().get(MimirRole.alertmanager, None)
        )
```

src/mimir_coordinator.py
```diff
@@ -163,6 +182,20 @@ def _build_compactor_config(self) -> Dict[str, Any]:
             "data_dir": str(self._root_data_dir / "data-compactor"),
         }
 
+    # ring.replication_factor: int
+    # Number of ingesters that each time series is replicated to. This option
+    # needs be set on ingesters, distributors, queriers and rulers when running in
+    # microservices mode.
+    def _build_ingester_config(self) -> Dict[str, Any]:
+        ingester_scale = len(self._cluster_provider.gather_addresses_by_role()[MimirRole.ingester])
```
Abuelodelanada, line 190: The same may occur here.

```suggestion
        ingester_scale = len(self._cluster_provider.gather_addresses_by_role().get(MimirRole.ingester, None))
```

src/mimir_coordinator.py
```diff
@@ -171,6 +204,22 @@ def _build_ruler_config(self) -> Dict[str, Any]:
...
         }
 
+    # sharding_ring.replication_factor:
+    # (advanced) The replication factor to use when sharding blocks. This option
+    # needs be set both on the store-gateway, querier and ruler when running in
+    # microservices mode.
+    def _build_store_gateway_config(self) -> Dict[str, Any]:
+        store_gateway_scale = len(
+            self._cluster_provider.gather_addresses_by_role()[MimirRole.store_gateway]
+        )
```
Abuelodelanada, line 214:
```suggestion
        store_gateway_scale = len(
            self._cluster_provider.gather_addresses_by_role().get(MimirRole.store_gateway, None)
        )
```

src/mimir_coordinator.py
lucabello: Done!

Abuelodelanada (approved):

PR #47, opened 2024-04-04
Self monitoring for the nginx in the coordinator.

Abuelodelanada:
This PR add self-monitoring capabilities for the Nginx running in the Coordinator charm using the nginx prometheus exporter

- Provide scrape job for nginx
- Provide alert rules for nginx

## How to test it.

Run the itests.


src/nginx.py
```diff
@@ -281,7 +301,9 @@ def _upstreams(self, addresses_by_role: Dict[str, Set[str]]) -> List[Dict[str, A
 
     def _locations(self, addresses_by_role: Dict[str, Set[str]]) -> List[Dict[str, Any]]:
         nginx_locations = []
+        nginx_locations.extend(LOCATIONS_BASIC)
```
sed-i, line 304:
```suggestion
        nginx_locations = LOCATIONS_BASIC.copy()
```

src/nginx_prometheus_exporter.py
```diff
@@ -0,0 +1,37 @@
...
+"""Nginx Prometheus exporter workload."""
+
+import logging
+
+from nginx import NGINX_PORT
+from ops.pebble import Layer
+
+logger = logging.getLogger(__name__)
+
+NGINX_PROMETHEUS_EXPORTER_PORT = "9113"
```
sed-i, line 12: This is unused in this file.
Is it possible to pass it as a cli arg to the exporter? Seems less brittle to have it explicit in the layer def.

src/prometheus_alert_rules/nginx/knyar-nginx-exporter.yml
sed-i, line 1: Should we pair this with a [dashboard](https://grafana.com/grafana/dashboards/?search=nginx)?

src/charm.py
```diff
@@ -134,7 +152,7 @@ def _on_config_changed(self, _: ops.ConfigChangedEvent):
 
     def _on_server_cert_changed(self, _):
         self._update_cert()
-        self._ensure_pebble_layer()
+        self._ensure_nginx_pebble_layer()
```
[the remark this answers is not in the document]
Abuelodelanada: good catch!
Besides `http` was fixed in `nginx_prometheus_exportes.py`

src/charm.py
```diff
@@ -267,7 +296,7 @@ def server_cert_path(self) -> Optional[str]:
     def internal_url(self) -> str:
         """Returns workload's FQDN. Used for ingress."""
         scheme = "https" if self._is_cert_available else "http"
-        return f"{scheme}://{socket.getfqdn()}:8080"
+        return f"{scheme}://{self.hostname}:8080"
```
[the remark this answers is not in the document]
Abuelodelanada: It's the same:

```python
    @property
    def hostname(self) -> str:
        """Unit's hostname."""
        return socket.getfqdn()
```

We are using `self.hostname` in other places too

src/charm.py
```diff
@@ -225,6 +246,14 @@ def workers_scrape_jobs(self) -> List[Dict[str, str]]:
             scrape_jobs.append(job)
         return scrape_jobs
 
+    @property
+    def coordinator_scrape_jobs(self) -> List[Dict[str, Any]]:
+        """Scrape jobs for the Mimir Coordinator."""
+        job: Dict[str, Any] = {
+            "static_configs": [{"targets": [f"{self.hostname}:{NGINX_PROMETHEUS_EXPORTER_PORT}"]}]
```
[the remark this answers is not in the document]
Abuelodelanada: It is not necessary, `metrics` is the default path if it is not present.

src/nginx_prometheus_exporter.py
Abuelodelanada: I placed it there since it is an nginx prometheus exporter default.
We only use it in charm.py.

We can use `--web.listen-address=:9113` in the exporter, but it is redundant...

src/prometheus_alert_rules/nginx/knyar-nginx-exporter.yml
Abuelodelanada: Yes, but I would address this in a separate issue since we need to make sure all the metrics the dashboard needs are there.

src/nginx.py
```diff
@@ -161,15 +163,43 @@
...
+    def __init__(self, charm: CharmBase, cluster_provider: MimirClusterProvider, server_name: str):
+        self._charm = charm
         self.cluster_provider = cluster_provider
         self.server_name = server_name
+        self._container = self._charm.unit.get_container("nginx")
+
+    def configure_pebble_layer(self) -> None:
+        """Configure pebble layer."""
+        self._container.push(
+            self.config_path, self.config(tls=self._charm._is_cert_available), make_dirs=True  # type: ignore
```
sed-i, line 199: Would it make sense to pass `_is_cert_available` as an arg to `configure_pebble_layer`?

src/prometheus_alert_rules/nginx/knyar-nginx-exporter.yml
```diff
@@ -0,0 +1,23 @@
+# Obtained from:
+# https://samber.github.io/awesome-prometheus-alerts/rules#nginx-1
+
+groups:
+- name: KnyarNginxExporter
```
sed-i, line 5: What does the `Knyar` prefix stand for?

src/nginx_prometheus_exporter.py
sed-i: I think specifying `f"--web.listen-address=:{NGINX_PROMETHEUS_EXPORTER_PORT}"` would be good practice because defaults may change (unlikely, but still), and we have code that relies on this.

sed-i (approved):

src/prometheus_alert_rules/nginx/knyar-nginx-exporter.yml
Abuelodelanada: It is the name of the author: https://samber.github.io/awesome-prometheus-alerts/rules#nginx

Abuelodelanada: Sure, I can it up a bit.
Anyway keep in mind that this is the name of the `group` not the `alert`

image

sed-i: Right! Never mind then :)

src/nginx.py
Abuelodelanada: Yes, it is better.
I will add a `_is_tls_ready` property... the `_is_cert_available` does not check the certs are in the container...

PR #59, opened 2024-06-14
fix: correctly expand meta-roles when gathering addresses

lucabello:
We apparently forgot to expand the meta roles, which we correctly do in the function above.


lucabello:
Irrelevant after the shared objects

PR #61, opened 2024-07-11
feat: implement generic coordinator

lucabello:
This PR is using the abstraction of the coordinator class in the Mimir charm.

This should **only** be merged after https://github.com/canonical/cos-lib/pull/31.

## Testing Instructions

CI will fail until the new cos-lib version is published; in the meantime, you can:
```bash
charmcraft pack
juju deploy ./mimir-coordinator.charm
# get the cos-lib file into the coordinator's container (via juju scp or anything)
# do the same for the worker
# try to use Mimir (note: you'll need to emulate s3 somehow)
```


lucabello:
Integration tests are not complete, but we're now in a better state than we had before (they were broken). I'd love to merge this PR, and open a new issue to make integration tests more complete.

lucabello:
https://github.com/canonical/mimir-coordinator-k8s-operator/issues/62

PR #65, opened 2024-08-02
fix roles and add tests for different deployment scenarios

lucabello:
We want to have integration tests for the following scenarios:
- one monolithic worker
- multiple workers, with scale 1
- multiple workers, with scale >1

This PR also fixes some issues with the Mimir roles found while testing.


src/mimir_config.py
```diff
@@ -17,12 +17,12 @@
 logger = logging.getLogger(__name__)
 
 ROLES = {
-    "overrides_exporter",
-    "query_scheduler",
+    "overrides-exporter",
```
PietroPasotti, line 20: ehm, given the PR description and title, it's not clear to me if this renaming change is intentional? and the alert rule?

tests/integration/helpers.py
```diff
@@ -65,3 +67,18 @@ async def get_unit_address(ops_test: OpsTest, app_name: str, unit_no: int):
     if unit is None:
         assert False, f"no unit exists in app {app_name} with index {unit_no}"
     return unit["address"]
+
+
+@retry(wait=wait_fixed(10), stop=stop_after_attempt(10))
+async def check_agent_data_in_mimir(ops_test: OpsTest, coordinator_app: str) -> Dict[str, Any]:
```
PietroPasotti, line 73: not a blocker, but I'm trying to make a habit of adding a comment (or a docstring) to explain what the test tests, unless obvious. In this case I'd have liked one. 
Also, adding something like: GIVEN/WHEN/THEN comments to spell out what you're doing and checking.

tests/integration/test_charm.py
```diff
@@ -90,37 +99,54 @@ async def test_metrics_endpoint(ops_test: OpsTest):
...
-    await ops_test.model.deploy("grafana-agent-k8s", "agent")
-
     await ops_test.model.integrate("mimir:mimir-cluster", "worker")
-    await ops_test.model.integrate("grafana:metrics-endpoint", "agent")
-    await ops_test.model.integrate("mimir:receive-remote-write", "agent")
-    await ops_test.model.integrate("mimir:s3", "s3")
-
     await ops_test.model.wait_for_idle(apps=["mimir", "worker", "agent", "s3"], status="active")
+    await check_agent_data_in_mimir(ops_test, "mimir")
+    await ops_test.model.remove_application(app_name="worker", destroy_storage=True)
```
PietroPasotti, line 113: I'm not very comfortable with doing model teardown stuff inside a test function. 
Try to follow this pattern https://discourse.charmhub.io/t/classify-tests-with-pytest-custom-markers-for-quick-integration-testing-iterations/14006

it's cleaner, and allows for much faster iterations when writing the tests

PietroPasotti (dismissed):
The test code needs some cleaning up, I recommend splitting it in several test files, one for each deployment scenario.

src/mimir_config.py
lucabello: It's intentional, I updated the PR description :)

lucabello:
We can split into one file per each; we'll triple the execution time though, is it really worth it?

tests/integration/helpers.py
lucabello: Added more documentation!

tests/integration/test_charm.py
lucabello: I'm splitting the tests, because all the workers using the same s3 backend isn't really testing much, as the data would be saved in one place only.

lucabello:
It's indeed worth it: having a shared S3 doesn't let us conclude anything on the three scenarios working (because the backend is still the same). I split the tests into three and marked the setup steps with `@pytest.mark.setup` so we can run only those with `-m setup`.

tests/integration/test_charm_ha_scaled.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@pytest.mark.setup
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest, mimir_charm: str):
+    """Build the charm-under-test and deploy it together with related charms."""
+    assert ops_test.model is not None  # for pyright
+    await ops_test.model.deploy(mimir_charm, "mimir", resources=charm_resources())
+
+    await ops_test.model.deploy("prometheus-k8s", "prometheus", channel="latest/edge")
```
PietroPasotti, line 33: you can group all consecutive 'deploy' calls in `asyncio.gather` fyi

PietroPasotti (approved):
lovely! thanks :)

PR #89, opened 2024-11-15
feat: support alert rules

### LINKED ISSUE #88: Support alert rules and alerting notifications in Mimir HA
### Enhancement Proposal

Currently, our Mimir HA solution doesn't support alerting. There are two main problems:

- the alert rules are sent over relation data to the coordinator, but we don't do anything with them;
- the internal alertmanager is not configured, nor configurable.

We should support alerting if we want Mimir to be a drop-in replacement for Prometheus, and fully functional in our COS bundle.

---

We agreed to use an external alertmanager (i.e., our `alertmanager-k8s` charm) instead of the internal Mimir alertmanager, because we still need an external one in the bundle for Loki.

lucabello:
## Issue

Closes #88.

## Solution

`mimirtool` is a binary from Grafana that can be pointed at a Mimir cluster, allowing us to easily set alert rules that are then persisted in s3. If the worker loses a pod, gets scaled to 0, if a new worker joins the cluster, the alert rules are propagated automatically by Mimir itself. `mimirtool` provides both a *holistic* and a *delta* approach; in this PR, I'm following what we do in Prometheus: replace holistically (regenerate the source of truth from relation data), but only if there are changes (compute the SHA and check before replacing).

The coordinator charm can gather alert rules via facilities from the `remote_write` library, and then save it to disk under `/etc/mimir-alerts/rules`. Those are then sent to the cluster via `mimirtool`, which is copied over from the charm container to the workload container.

This PR also adds the integration with alertmanager, allowing to `juju relate` the coordinator to it, and correctly configuring the alertmanager urls in Mimir so that the alerts go through correctly.

### Passing fixes

The meta-roles definition was wrong: the role `all` did not, in fact, include **all roles**, but only a subset. I adjusted the recommended deployment accordingly. I also removed `alertmanager` from the recommended deployment, as we decided to use the external one.

## Testing Instructions

Deploy the Mimir HA solution with the coordinator from this PR. Also deploy `avalanche-k8s` and `alertmanager-k8s`, and relate them to the coordinator.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -140,7 +140,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
 
     def _on_pebble_ready(self, _) -> None:
         """Make sure the `mimirtool` binary is in the workload container."""
-        self._push_mimirtool()
+        self._ensure_mimirtool()
 
     ######################
     # === PROPERTIES === #
@@ -211,8 +211,10 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
 
         return paths
 
-    def _push_mimirtool(self):
+    def _ensure_mimirtool(self):
         """Copy the `mimirtool` binary to the workload container."""
+        if self._nginx_container.exists("/usr/bin/mimirtool"):
+            return
         with open("mimirtool", "rb") as f:
             self._nginx_container.push("/usr/bin/mimirtool", source=f, permissions=0o744)
 
@@ -226,8 +228,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
             return hashlib.sha256(hashable).hexdigest()
 
         # Get mimirtool if this is the first execution
-        if not self._pull(ALERTS_HASH_PATH):
-            self._push_mimirtool()
+        self._ensure_mimirtool()
 
         remote_write_alerts = self.remote_write_provider.alerts
         alerts_hash = sha256(str(remote_write_alerts))
@@ -253,7 +254,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
             if mimirtool_output.stdout:
                 logger.info(f"mimirtool: {mimirtool_output.stdout.read().strip()}")
             if mimirtool_output.stderr:
-                logger.info(f"mimirtool (err): {mimirtool_output.stderr.read().strip()}")
+                logger.error(f"mimirtool (err): {mimirtool_output.stderr.read().strip()}")
 
 
 if __name__ == "__main__":  # pragma: nocover


lucabello:
Everything has been tested manually: you can see the alerts appearing in the Mimir datasource in Grafana:

Screenshot-2024-11-15_12-27

The relation with alertmanagere is also working as intended:
Screenshot-2024-11-15_11-18

charmcraft.yaml
```diff
@@ -27,3 +27,12 @@ parts:
     override-pull: |
       curl -L -O https://github.com/canonical/cos-tool/releases/latest/download/cos-tool-${CRAFT_TARGET_ARCH}
       chmod +x cos-tool-*
+  mimirtool:
+    plugin: dump
+    source: .
+    build-packages:
+      - curl
+    override-pull: |
+      curl -fLo mimirtool https://github.com/grafana/mimir/releases/latest/download/mimirtool-linux-${CRAFT_TARGET_ARCH}
```
sed-i, line 36: Is mimir tool released together with mimir? Should we be taking the mimirtool that matches the oci image? Is there an easy way to make sure we match? This can be a separate issue.

src/charm.py
```diff
@@ -132,6 +137,10 @@ def _on_ingress_revoked(self, _) -> None:
         """
         logger.info("Ingress for app revoked")
 
+    def _on_pebble_ready(self, _) -> None:
+        """Make sure the `mimirtool` binary is in the workload container."""
+        self._get_mimirtool()
```
sed-i, line 142: push_mimirtool?

src/charm.py
```diff
@@ -165,12 +174,88 @@ def external_url(self) -> str:
...
+        """Create alert rule files for all Mimir consumers."""
+
+        def sha256(hashable: Any) -> str:
+            """Use instead of the builtin hash() for repeatable values."""
+            if isinstance(hashable, str):
+                hashable = hashable.encode("utf-8")
+            return hashlib.sha256(hashable).hexdigest()
+
+        if not self._nginx_container.can_connect():
+            return
```
sed-i, line 228: Should we drop this guard?
- It's point in time anyway
- Caller should guard anyway
- In some cases, a guard here can get us into model departure

src/charm.py
```diff
@@ -165,12 +174,88 @@ def external_url(self) -> str:
...
+
+        # Get mimirtool if this is the first execution
+        if not self._pull(ALERTS_HASH_PATH):
+            self._get_mimirtool()
```
sed-i, line 232: Seems odd to push mimirtool if no alert hash. The two are technically unrelated.
(mimirtool was already pushed on pebble ready.)
Am I missing something?

src/charm.py
```diff
@@ -165,12 +174,88 @@ def external_url(self) -> str:
...
+                    *rules_file_paths,
+                    f"--address={self.external_url}",
+                    "--id=anonymous",  # multitenancy is disabled, the default tenant is 'anonymous'
+                ],
+                encoding="utf-8",
+            )
+            if mimirtool_output.stdout:
+                logger.info(f"mimirtool: {mimirtool_output.stdout.read().strip()}")
+            if mimirtool_output.stderr:
+                logger.info(f"lokitool (err): {mimirtool_output.stderr.read().strip()}")
```
sed-i, line 258:
```suggestion
                logger.info(f"mimirtool (err): {mimirtool_output.stderr.read().strip()}")
```

src/charm.py
```diff
@@ -165,12 +174,88 @@ def external_url(self) -> str:
...
+        if not self._nginx_container.can_connect():
+            return
+
+        # Get mimirtool if this is the first execution
+        if not self._pull(ALERTS_HASH_PATH):
+            self._get_mimirtool()
+
+        remote_write_alerts = self.remote_write_provider.alerts
+        alerts_hash = sha256(str(remote_write_alerts))
+        alert_rules_changed = alerts_hash != self._pull(ALERTS_HASH_PATH)
```
sed-i, line 236: - We already pulled once a few lines above. Could use walrus above.
- Using the hash has a nice side effect: manually deleting alerts won't trigger recreation; only deleting the hash file would. Should we have an issue to add this to a troubleshooting doc?

src/mimir_config.py
```diff
@@ -37,17 +37,25 @@
 """Mimir component role names."""
 
 META_ROLES = {
-    "read": {"query-frontend", "querier"},
+    "read": {"querier", "query-frontend"},
     "write": {"distributor", "ingester"},
     "backend": {
+        "alertmanager",
```
sed-i, line 43: `alertmanager` is part of `backend` but not part of `all`?

src/mimir_config.py
```diff
@@ -57,26 +65,24 @@
     "ingester",
     "querier",
     "query-frontend",
-    "query-scheduler",
+    # "query-scheduler",
```
sed-i, line 68: Add comment explaining why this is commented out?

charmcraft.yaml
lucabello: They are released together; I'm not sure how we would do that, because the Mimir image version is only known at runtime.We can manually pick a version on the same major as the image we

src/charm.py
lucabello: `self._set_alerts()` is called in the constructor, so I believe it can be called before `_on_pebble_ready`. That `if` is meant to make sure we always have `mimirtool` on the first execution.

Is there a better way to do this?

src/charm.py
lucabello: Fixed

src/charm.py
lucabello: Created in Jira!

src/charm.py
```diff
@@ -165,12 +175,85 @@ def external_url(self) -> str:
...
+            self._push(path, rules)
+            paths.append(path)
+            logger.debug("Updated alert rules file %s", filename)
+
+        return paths
+
+    def _push_mimirtool(self):
+        """Copy the `mimirtool` binary to the workload container."""
+        with open("mimirtool", "rb") as f:
+            self._nginx_container.push("/usr/bin/mimirtool", source=f, permissions=0o744)
```
Abuelodelanada, line 217: Shouldn't we verify whether mimirtool is already in the container or not to avoid transferring ~75Mb each time a `pebble_ready` event is fired or `self._push_mimirtool()` is executed? 

I would name this method `_ensure_mimirtool` or something like this...

src/charm.py
```diff
@@ -165,12 +175,85 @@ def external_url(self) -> str:
...
+                    *rules_file_paths,
+                    f"--address={self.external_url}",
+                    "--id=anonymous",  # multitenancy is disabled, the default tenant is 'anonymous'
+                ],
+                encoding="utf-8",
+            )
+            if mimirtool_output.stdout:
+                logger.info(f"mimirtool: {mimirtool_output.stdout.read().strip()}")
+            if mimirtool_output.stderr:
+                logger.info(f"mimirtool (err): {mimirtool_output.stderr.read().strip()}")
```
Abuelodelanada, line 256: If there is an error in the mimirtool execution... shouldn't we log an error?

```suggestion
            if mimirtool_output.stderr:
                logger.error(f"mimirtool (err): {mimirtool_output.stderr.read().strip()}")
```

Is this enough? 🤔

src/charm.py
```diff
@@ -165,12 +175,85 @@ def external_url(self) -> str:
...
+
+        def sha256(hashable: Any) -> str:
+            """Use instead of the builtin hash() for repeatable values."""
+            if isinstance(hashable, str):
+                hashable = hashable.encode("utf-8")
+            return hashlib.sha256(hashable).hexdigest()
+
+        # Get mimirtool if this is the first execution
+        if not self._pull(ALERTS_HASH_PATH):
+            self._push_mimirtool()
```
sed-i, line 230: Why not check instead if the mimir file exists?

src/charm.py
lucabello: Good point! Yes this should be enough :)

Abuelodelanada (approved):

Abuelodelanada (approved):
LGTM!

PR #90, opened 2024-11-21
Migrate away from Tox to UV with Make

MichaelThamm:
> [!IMPORTANT]
>  Switch CI branch to main once tandem PR is merged:
- https://github.com/canonical/observability/pull/229

Switching to UV and Make in favour of Tox due to speed and a long-term plan to improve CI. To test with this PR you can run the following commands:

* `tox -e lint` -> `make unit`
* `tox -e static` -> `make static`
* `tox -e fmt` -> `make fmt`
* `tox -e unit` -> `make unit`
* `tox -e scenario` -> `make scenario`
* `tox -e integration` -> `make integration`

Want to use args?
* `make integration ARGS=--keep-models `

Currently missing:
- [x] scenario tests
- [x] colour formatting in GH CI output
- [x] requirements.txt management
- [x] Can we consolidate the charmcraft.yaml `charm-binary-python-packages` since we have a pyproject.toml (single source of truth)

CI for `charm-pull-request.yaml` is being updated to conditionally test with make instead of tox depending on the existence of a `tox.ini` file in the repo root.

Tandem PR:
- https://github.com/canonical/mimir-worker-k8s-operator/pull/84


charmcraft.yaml
```diff
@@ -12,6 +12,13 @@ bases:
 
 parts:
   charm:
+    build-snaps:
+      - astral-uv
+    
+    override-build: |
+      make generate-requirements
```
MichaelThamm, line 19: This change creates the requirements.txt on charmcraft pack or building the charm. I noticed that when I run `ARGS=--keep-models make integration` and I `juju debug-log --include mimir` then I get the error that `cannot import module crossplane` which was specified in the `pyproject.toml`.

1. When I `juju ssh mimir/0 cat /var/lib/juju/agents/unit-mimir-0/charm/requirements.txt | grep crossplane` I can see that crossplane exists
2. When I `juju ssh mimir/0 ls /var/lib/juju/agents/unit-mimir-0/charm/venv | grep crossplane` I can see that crossplane does not exist.

This means that the make command in the charmcraft.yaml works in creating the requirements.txt but its contents are not respected.

Makefile
```diff
@@ -0,0 +1,66 @@
...
+	uv tool run ruff check --fix-only $(ALL)
+	uv tool run ruff format $(ALL)
+
+unit:
+	uv run --isolated \
+	    --extra unit \
+		coverage run \
+		--source=$(SRC) \
+		-m pytest \
+		--ignore=$(TESTS)/integration \
```
ca-scribner, line 36: `unit` tests everything in this repo, but then uses `--ignore` to omit the integration tests.  scenario and integration test only specific folders.  We should probably use just one of those two methods

also, I think `unit` test right now will run sidecar tests too

charmcraft.yaml
ca-scribner: I dont know why this would occur, but just in case have you tried doing a `charmcraft clean` before packing?  maybe there's something leftover in the cache

Makefile
MichaelThamm: Maybe there is a typo in your comment or I do not fully understand, do you mean to remove the --ignore because it is the same as this conversation:

https://github.com/canonical/tempo-coordinator-k8s-operator/pull/62#discussion_r1842576307

charmcraft.yaml
MichaelThamm: Same result unfortunately, will have to investigate some more.

Makefile
```diff
@@ -0,0 +1,67 @@
...
+		--tb native \
+		-v \
+		-s \
+		$(ARGS)
+	uv run --extra unit coverage report
+
+scenario:
+	@echo "Add scenario tests here ..."
+	@# uv run --isolated \
```
MichaelThamm, line 45: In CI we can check if we have a scenario folder and conditionally run make/tox scenario if exists.

Makefile
```diff
@@ -0,0 +1,60 @@
...
+export PYTHONPATH = $(PROJECT):$(PROJECT)/lib:$(SRC)
+export PY_COLORS=1
+
+# Update uv.lock with the latest deps
+lock:
+	uv lock --upgrade --no-cache
+
+# Generate requirements.txt from pyproject.toml
+requirements:
+	uv export --frozen --no-hashes --format=requirements-txt -o requirements.txt
```
lucabello, line 16: Have we figured out why output redirection didn't work in the `charmcraft pack` context? I'm concerned that might *make* us lose some time in the future

MichaelThamm: No solution yet ...

Abuelodelanada:
`tox -e lint -> make unit` is this ok?

shouldn't be: `tox -e lint -> make lint`?

Abuelodelanada (approved):

PR #100, opened 2024-12-13
Implement grafana_datasource_exchange

michaeldmitry:
## Issue
Add a `send-datasource` endpoint to exchange datasource UIDs with other COS components.

## Testing Instructions
1. Deploy Mimir HA
2. Deploy `grafana-k8s`
3. Deploy Tempo HA
4. Integrate `tempo` and `grafana` over `grafana_datasource`
5. Integrate `mimir` and `grafana` over `grafana_datasource`
6. Integrate `mimir` and `tempo` over `grafana_datasource_exchange`
7. `jhack show-relation mimir tempo` and observe the datasource UID of `mimir` shared to `tempo` 

## Drive-bys
- Add interface and scenario tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,12 +44,11 @@ unit:
 		--capture=no \
 		$(TESTS)/unit \
 		$(ARGS)
-	uv run --frozen --isolated --extra dev \
-		coverage report
 	
 	uv run --frozen --isolated --extra dev \
 		coverage run \
 		--source=$(SRC) \
+		--append \
 		-m pytest \
 		--tb native \
 		--verbose \
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ fmt:
 	uv run --frozen --isolated --extra dev \
 		ruff format $(ALL)
 
-# Run unit and scenario tests
+# Run unit tests
 unit:
 	uv run --frozen --isolated --extra dev \
 		coverage run \
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ fmt:
 	uv run --frozen --isolated --extra dev \
 		ruff format $(ALL)
 
-# Run unit tests
+# Run unit and scenario tests
 unit:
 	uv run --frozen --isolated --extra dev \
 		coverage run \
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -58,3 +58,27 @@ integration:
 		--log-cli-level=INFO \
 		$(TESTS)/integration \
 		$(ARGS)
+
+# Run scenario tests
+scenario:
```
PietroPasotti, line 63: I think that we should start collapsing unit and scenario tests into unit.

src/charm.py
```diff
@@ -113,6 +116,9 @@ def __init__(self, *args: Any):
         if self._nginx_container.can_connect():
             self._set_alerts()
 
+        # do this regardless of what event we are processing
+        self._reconcile()
```
PietroPasotti, line 120: while I agree that this is a good way to go, it makes me wonder why we didn't do the same in prometheus and loki non-HA and instead you chose to explicitly observe four different events?
Also there's more unconditional logic I see just above this line, that should be moved into _reconcile, but that feels like a refactoring that's easily out of scope for this PR. We should ask @lucabello or mike if they're fine with this change

michaeldmitry: The reason I chose not to follow this pattern in non-ha charms was that it seemed that it'd need a whole lot of refactoring for the rest of the charm to follow the reconcile, so might as well just stick to what's already there

Makefile
michaeldmitry: combined them both under `unit` in the Makefile

src/charm.py
michaeldmitry: for this charm, I refactored it to follow this reconcile pattern

src/charm.py
```diff
@@ -256,6 +230,33 @@ def sha256(hashable: Any) -> str:
...
+                )
+        self.coordinator.datasource_exchange.publish(datasources=raw_datasources)
+
+    def _reconcile(self):
+        # This method contains unconditional update logic, i.e. logic that should be executed
+        # regardless of the event we are processing.
+        if self._nginx_container.can_connect():
+            self._set_alerts()
+        self._ensure_mimirtool()
+        self._update_datasource_exchange()
```
Abuelodelanada, line 258: There is some information we've lost with this re-factor, for instance:

```python
logger.info("Ingress for app ready on '%s'", event.url)
```

Despite of the fact I like this pattern because makes everything clean, we need to come up with a standardised way to process information which is only available in the specific event.

michaeldmitry: I was thinking that the intent behind observing `self.ingress.on.ready` and `self.ingress.on.revoked` was to wake up the charm (for refresh purposes), which is exactly what a common exit hook would do.
If we need that log message, I can keep the observers to log while maintaing the common logic like 
```
if self._nginx_container.can_connect():
            self._set_alerts()
self._ensure_lokitool()
```
inside `_reconcile`

src/charm.py
PietroPasotti: Yeah I think it's a good idea not to mix the two in charms that are too old-school

PietroPasotti (approved):
LGTM!

PR #114, opened 2025-03-03
refactor: change MimirCoordinatorK8sOperatorCharm.external_url, add most_external_url

ca-scribner:
## Issue
Previously, MimirCoordinatorK8sOperatorCharm.external_url returned the external url if it was available, otherwise it returned the ingress url.  This meant that any consuming code that explicitly wanted the external_url or nothing needed to duplicate checks that exist in external_url, making it easy to later have these go out of sync.

## Solution
This change breaks external_url into two functions:
* external_url, which explicitly returns only an external url or None (if it isn't available)
* most_external_url, which returns the external_url if it is available, otherwise the internal url

This is useful for a planned future PR which will need exactly the external_url

## Context
This is a precursor PR to then implement the prometheus-api library, which will send out mimir's ingress and direct prometheus-api url.

## Testing Instructions
Existing tests cover this change

## Upgrade Notes
No outward facing changes included

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -67,7 +67,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
         self.coordinator = Coordinator(
             charm=self,
             roles_config=MIMIR_ROLES_CONFIG,
-            external_url=self.external_url,
+            external_url=self.most_external_url,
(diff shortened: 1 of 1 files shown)


ca-scribner:
This ended up getting merged as part of #115

PR #115, opened 2025-03-03
feat: add prometheus-api lib, relation

ca-scribner:
## Issue
The Mimir coordinator charm does not provide a way for other charms to consume its prometheus api for querying metrics.  

## Solution
Implemented here is a library that sends Mimir's prometheus api url to other applications, plus tests.  Schema for the data sent is:

```python
class PrometheusApiAppData(BaseModel):
    """Data model for the prometheus-api interface."""

    ingress_url: Optional[AnyHttpUrlOrStrUrl] = Field(
        default=None,
        description="The non-internal URL at which this application can be reached.  Typically, this is an ingress URL.",

    )
    direct_url: AnyHttpUrlOrStrUrl = Field(
        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
                    " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
                    " from inside the cluster, with scheme."
    )
```

This will enable connecting Mimir to other charms such as Kiali.

## Context
This implementation is essentially identical to that of [grafana-metadata](https://github.com/canonical/grafana-k8s-operator/pull/388).

Merge #114 before merging this.  

## Testing Instructions
Automated tests cover everything changed here

## Upgrade Notes
No upgrade considerations

### DIFF
--- a/lib/charms/mimir_coordinator_k8s/v0/prometheus_api.py
+++ b/lib/charms/mimir_coordinator_k8s/v0/prometheus_api.py
@@ -1,0 +1,226 @@
+"""prometheus_api.
+
+This library implements endpoint wrappers for the prometheus-api interface.  The prometheus-api interface is used to
+transfer information about an instance of Prometheus, such as how to access and uniquely identify it.  Typically, this is
+useful for charms that operate a Prometheus instance to give other applications access to its API.
+
+## Usage
+
+### Requirer
+
+PrometheusApiRequirer is a wrapper for pulling data from the prometheus-api interface.  To use it in your charm:
+
+* observe the relation-changed event for this relation wherever your charm needs to use this data (this endpoint wrapper
+  DOES NOT automatically observe any events)
+* wherever you need access to the data, call `PrometheusApiRequirer(...).get_data()`
+
+An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+
+        self.prometheus_api = PrometheusApiRequirer(self.model.relations, "prometheus-api")
+
+        self.framework.observe(self.on["prometheus-api"].relation_changed, self._on_prometheus_api_changed)
+
+    def _on_prometheus_api_changed(self):
+        data = self.prometheus_api.get_data()
+        ...
+```
+
+Where you also add relation to your `charmcraft.yaml` or `metadata.yaml` (note that PrometheusApiRequirer is designed
+for relating to a single application and must be used with limit=1 as shown below):
+
+```yaml
+requires:
+  prometheus-api:
+    limit: 1
+    interface: prometheus_api
+```
+
+### Provider
+
+PrometheusApiProvider is a wrapper for publishing data to charms related using the prometheus-api interface.  Note
+that `PrometheusApiProvider` *does not* manage any events, but instead provides a `publish` method for sending data to
+all related applications.  Triggering `publish` appropriately is left to the charm author, although generally you want
+to do this at least during the `relation_joined` and `leader_elected` events.  An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.prometheus_api = PrometheusApiProvider(
+            relations=self.model.relations,
+            relation_name="prometheus-api",
+            app=self.app,
+        )
+
+        self.framework.observe(self.on.leader_elected, self.do_something_to_publish)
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -271,6 +287,21 @@ def sha256(hashable: Any) -> str:
...
+            return
+
+        prometheus_api = PrometheusApiProvider(
+            relation_mapping=self.model.relations,
+            app=self.app,
+            relation_name=prometheus_api_relation_name,
+        )
+        prometheus_api.publish(
+            direct_url=self.internal_url,
+            ingress_url=self.external_url,
```
ca-scribner, line 302: Help wanted here please :)  

Are these the correct URLs to hit the prometheus-compliant api that mimir provides?  I suspect they aren't, but I wasn't sure what was the correct thing to put here :)

simskij: They aren't :) If i recall correctly it's `/prometheus/`, but you kind of need to verify that before merging.

ca-scribner: I don't see anywhere using `/prometheus`, but I see the remote write adding [``/api/v1/push`](https://github.com/canonical/mimir-coordinator-k8s-operator/blob/744af60c99abe6b4e412e24be82a6d6829a9ec45/src/charm.py#L115).  So makes me think this should include `/api` or `/api/v1`

lib/charms/mimir_coordinator_k8s/v0/prometheus_api.py
```diff
@@ -0,0 +1,226 @@
...
+    ingress_url: Optional[AnyHttpUrlOrStrUrl] = Field(
+        default=None,
+        description="The non-internal URL at which this application can be reached.  Typically, this is an ingress URL.",
+
+    )
+    direct_url: AnyHttpUrlOrStrUrl = Field(
+        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
+                    " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
+                    " from inside the cluster, with scheme."
+    )
```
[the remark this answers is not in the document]
ca-scribner: the owning charm is mimir, but the **api** is prometheus.  iiuc, Mimir provides a prometheus-compliant APi

direct_url is for the application (`The cluster-internal URL at which this application can be reached`)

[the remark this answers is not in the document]
ca-scribner: For the prometheus scaling, if that's something that blocks for you lets discuss.  I don't see the value (is anyone using prometheus scaled, and do we want them to?)

I see a good argument against ingress-per-unit because then the prometheus charm provides multiple independent prometheuses and the developer of the consuming app gets to choose.  I don't see how that's a UX that works for an administrator.  imo the only real way is to do ingress-per-app, the leader is the current prometheus, and any scaled units are backups that could step in

src/charm.py
ca-scribner: I'm a liar - I see `/prometheus` in the grafana source [here](https://github.com/canonical/mimir-coordinator-k8s-operator/blob/744af60c99abe6b4e412e24be82a6d6829a9ec45/src/charm.py#L102).  So I'm not 100% sure what should be used :D I'll have to do some test queries

sed-i (approved):
Discussed with @ca-scribner, raised some drawbacks to be potentially addressed in the future:
- traefik - nginx - unit addresses
- what do we report for when the coordinator is scaled up - usage of leader guard
- why roll responsibility of "remote-model-get" to the requirer if the provider should already know, i.e. why the provider can't report just one url.

Approving with the reservations above in mind.

src/charm.py
ca-scribner: As always, I shouldn't doubt @simskij :)

I think what confused me was that:
* prometheus by default (or maybe just always?) serves its api at its top level path.  eg: get labels at `prometheus_ip:port/api/v1/labels`
* mimir by default serves its prometheus api at `/prometheus`, eg: get labels at `mimir_ip:port/prometheus/api/v1/labels`.  [this doc](https://grafana.com/docs/mimir/latest/configure/configure-the-query-frontend-work-with-prometheus/) mentions how the path is configurable, though I can't find where the full query-frontend configuration schema is documented

To start, I can update this url to be `/prometheus`, but leaving it hard coded here feels like an easy accidental bug if someone ever exposed the query-frontend `api.prometheus_http_prefix` config.  Maybe we should explicitly set it in the mimir charm just so we never forget?

ca-scribner: For now, this is hard coded in a few other places so I've done the same.

PR #120, opened 2025-03-20
fix: missing `/prometheus` in prometheus-api integration, other minor fixes

ca-scribner:
## Issue
#115 implemented the prometheus-api relation but exposed the wrong url for the prometheus api.  Mimir's prometheus api is located at `/prometheus`, not the root path

## Solution
Update the url sent over prometheus-api

## Context
#115

## Testing Instructions
Covered by unit tests.  Can also test this by doing prometheus queries on COORDINATOR_IP/prometheus

## Upgrade Notes
-


ca-scribner:
Closed in favor of #123

PR #124, opened 2025-04-01
Bring back `tox`

michaeldmitry:
## Issue
TICS scan [fails](https://github.com/canonical/mimir-coordinator-k8s-operator/actions/runs/14161243002/job/39666975392) due to missing `tox.ini` file 

## Solution
Instead of accommodating `make` commands in the TICS workflow, we, anyway, need to bring back tox.


lucabello:
Superseded by #125

Issue #128, opened 2025-04-29
Mimir errors out on Canonical K8s due to hardcoded DNS resolver config


vishvikkrishnan:
### Bug Description

Because of hardcoded resolver name in the [nginx config](https://github.com/canonical/mimir-coordinator-k8s-operator/blob/main/src/nginx_config.py#L318) of the charm code, the mimir units fail on Canonical K8s where the DNS is coredns instead of kube-dns.

This bug prevents deployment of COS-HA on top of Canonical K8s.

For reference, the [nginx config for tempo](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/nginx_config.py#L205) does not suffer from this issue since the DNS resolver address is fetched from `/etc/resolv.conf` instead of a hard-coded resolver name.

### To Reproduce

1. Ensure juju model is available on a **Canonical K8s** cloud (it's important that the K8s cloud is based on Canonical K8s, not MicroK8s).
2. Deploy COS-HA using the terraform modules in the [observability repo](https://github.com/canonical/observability/blob/main/terraform/modules/cos/README.md#usage).
3. Wait until mimir triggers the `receive-remote-write-relation-changed` hook.

### Environment

juju version 3.6.4
canonical k8s version: 1.32.2
mimir from latest/edge, revision 40
(all other cos components version are decided by the terraform module linked in the "To Reproduce" section)

### Relevant log output

```shell
unit-mimir-0: 18:24:38 ERROR juju.worker.uniter.operation hook "receive-remote-write-relation-changed" (via hook dispatching script: dispatch) failed: exit status 1
unit-mimir-0: 18:24:38 INFO juju.worker.uniter awaiting error resolution for "relation-changed" hook
unit-mimir-0: 18:24:45 INFO juju.worker.uniter awaiting error resolution for "relation-changed" hook
unit-mimir-0: 18:29:38 INFO juju.worker.uniter awaiting error resolution for "relation-changed" hook
unit-mimir-0: 18:29:38 WARNING unit.mimir/0.receive-remote-write-relation-changed this module has been deprecated and may be removed in a future version: please use cosl.interfaces.cluster and cosl.interfaces.utils instead.
unit-mimir-0: 18:29:44 ERROR unit.mimir/0.juju-log receive-remote-write:57: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-mimir-0/charm/src/charm.py", line 342, in <module>
[...]

PR #130, opened 2025-05-06
Simplify nginx config generation

michaeldmitry:
## Issue
All our current three coordinators generate their own nginx config from scratch although they share a lot of common config.

### Drive-by
Resolves https://github.com/canonical/mimir-coordinator-k8s-operator/issues/128

## Solution
Extract all shared configuration into `cosl`, and keep only the coordinator-specific parts of the `upstreams` and `locations` blocks, which each coordinator will pass to the shared object as needed.

## Context
https://github.com/canonical/cos-lib/pull/143

## Drive-bys
- Update lock file


MichaelThamm (approved):
Looks great, thanks Michael

Issue #141, opened 2025-06-03
Investigate the GRPC communication in Mimir and how to secure it with TLS


lucabello:
While addressing https://github.com/canonical/mimir-worker-k8s-operator/issues/104 , removing the `grpc_tls_config` solves the issue. However, we felt the need to investigate and reach a better understanding of what's going on behind the scenes.

lucabello:
If we configure a `grpc_client` for most Mimir components, at some point we see that Mimir is trying to contact `127.0.0.1:9095` over GRPC - but the certificate isn't valid for that address (ít should be using the hostname, not the localhost ip). While configuring the `grpc_listen_port` does impact the ip we're trying to reach, changing the `grpc_listen_address` doesn't.

Removing the `grpc_*` configurations yields a working cluster.
Looking closely at the `pebble logs` during the startup sequence (or a `pebble restart`), we can see that Mimir detects "monolithic mode" (role `all`, but TODO does it happen even when scaled?) and configures query-frontend and query-scheduler to be reachable at `127.0.0.1:9095`[^1].



[^1]: Mimir uses port `9095` both externally and for internal components communication over grpc: https://github.com/grafana/mimir/discussions/7047

PR #142, opened 2025-06-04
fix: remove broken grpc tls section

lucabello:
## Issue
Closes https://github.com/canonical/mimir-worker-k8s-operator/issues/104

## Solution
The GRPC section causes issues and prevents from Mimir from starting. I wasn't able to achieve a working configuration for that section, so we're dropping TLS for GRPC - we never had it working, and it won't matter once we're on a service mesh.

## Testing Instructions
Deploy and test things as normal, but with self-signed-certificate :D


Abuelodelanada (approved):
easy fix!

src/mimir_config.py
```diff
@@ -162,7 +162,6 @@ def _build_tls_config(self) -> Dict[str, Any]:
         }
         return {
             "http_tls_config": tls_config,
-            "grpc_tls_config": tls_config,
```
sed-i, line 165: Please add a FIXME with link to [issue](https://github.com/canonical/mimir-coordinator-k8s-operator/issues/141).
If we can also keep a blank placeholder, even better:
```
"grpc_tls_config": {
    # FIXME ...
},
```

tests/unit/test_mimir_config.py
```diff
@@ -200,12 +200,6 @@ def test_build_tls_config(mimir_config):
             "client_ca_file": "/etc/worker/ca.cert",
             "client_auth_type": "RequestClientCert",
         },
-        "grpc_tls_config": {
-            "cert_file": "/etc/worker/server.cert",
-            "key_file": "/etc/worker/private.key",
-            "client_ca_file": "/etc/worker/ca.cert",
-            "client_auth_type": "RequestClientCert",
-        },
```
sed-i, line 208: Similar to the comment above - please add a FIXME.

PR #149, opened 2025-06-18
fix: disable the default max alerts per tenant to work with cos

### LINKED ISSUE #148: When too many alert rules are present they aren't synced at all
### Bug Description

During PC testing, I noticed that alert rules were not making it to the Mimir storage (i.e., they were not being `mimirtool`-synced).

Upon further inspection, I found all alerts in relation data, and also on disk in the coordinator.

The problem is `mimirtool sync` is failing silently. I `ssh`-ed to the nginx container in the coordinator, and tried to run that `mimirtool sync` command manually, which yields a BadRequest error:
```
per-user rule groups limit (limit: 70 actual: 77) exceeded
```

The aggregated alert rules for the entire COS deployment are saved to a single file, because they're all coming from grafana-agent. There is a config option we can tweak in the [`limits:`](https://grafana.com/docs/mimir/latest/configure/configuration-parameters/#limits) section:

```yaml
# Maximum number of rule groups per-tenant. 0 to disable.
# CLI flag: -ruler.max-rule-groups-per-tenant
[ruler_max_rule_groups_per_tenant: <int> | default = 70]
```

We should either increase the default to something sensible, or expose this config as a config option

### To Reproduce

Deploy Mimir and enough stuff so that there's more than 70 alert rules.

Alternatively, deploy Mimir and ask AI to give you 71 alert rules.

### Environment

- mimir: rev53 (`2/edge`)
- mimir-worker: rev55 (`2/edge`)

### Relevant log output

```shell
ERRO[0000] error during request to Grafana Mimir API     error="Post \"http://10.19.2.2/cos-2-edge-mimir/prometheus/config/v1/rules/juju_cos-2-edge_44a95acd_grafana-agent\": POST request to http://10.19.2.2/cos-2-edge-mimir/prometheus/config/v1/rules/juju_cos-2-edge_44a95acd_grafana-agent failed: server returned HTTP status: 400 Bad Request, body: \"per-user rule groups limit (limit: 70 actual: 77) exceeded\\n\"" method=POST url="http://10.19.2.2/cos-2-edge-mimir/prometheus/config/v1/rules/juju_cos-2-edge_44a95acd_grafana-agent"
```

### Additional context

_No response_

lucabello:
## Issue
Closes #148.

## Solution
Disable the per-tenant alert limit.

## Context
This was discovered and addressed in the context of Partner Cloud testing.

## Testing Instructions
Applying the COS terraform module and then refreshing the charm to include these local changes. I manually ran this test in the Partner Cloud deployment.

### DIFF
--- a/src/mimir_config.py
+++ b/src/mimir_config.py
@@ -315,6 +315,6 @@ class MimirConfig:
     # Maximum number of rule groups per-tenant. 0 to disable. (default 70)
     def _build_limits_config(self) -> Dict[str, Any]:
         return {
-            "ruler_max_rules_per_rule_group": 100,
-            "ruler_max_rule_groups_per_tenant": 200,
+            "ruler_max_rules_per_rule_group": 0,
+            "ruler_max_rule_groups_per_tenant": 0,
         }
--- a/src/mimir_config.py
+++ b/src/mimir_config.py
@@ -137,6 +137,7 @@ class MimirConfig:
             "store_gateway": self._build_store_gateway_config(coordinator.cluster),
             "blocks_storage": self._build_blocks_storage_config(),
             "memberlist": self._build_memberlist_config(coordinator.topology, coordinator.cluster),
+            "limits": self._build_limits_config(),
         }
 
         if coordinator.s3_ready:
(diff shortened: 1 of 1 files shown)


PietroPasotti (commented):
the code looks good (although: no tests? how do we test this?)

but I cannot make the call as to whether this is the right way to do it.
Do we need to surface those as config options?
Do we need to set a higher default, instead of disabling it altogether?

PR #153, opened 2025-06-24
feat: add exemplar storage with default value

sinapah:

### DIFF
--- a/src/mimir_config.py
+++ b/src/mimir_config.py
@@ -317,4 +317,5 @@ class MimirConfig:
         return {
             "ruler_max_rules_per_rule_group": 0,
             "ruler_max_rule_groups_per_tenant": 0,
+            "max_global_exemplars_per_user": 100000
         }


src/mimir_config.py
```diff
@@ -116,10 +116,12 @@ def __init__(
         alertmanager_urls: Set[str] = set(),
         root_data_dir: Path = Path("/data"),
         recovery_data_dir: Path = Path("/recovery-data"),
+        max_global_exemplars_per_user: int = 100000
```
sed-i, line 119: Do we need to have a default both here and in charmcraft.yaml? I.e. perhaps `Optional[int]`?

How about the following tests:

```python
# WHEN arg is omitted
# THEN rendered config does not have a `max_global_exemplars_per_user` entry

# WHEN arg is 0
# THEN rendered config does not have a `max_global_exemplars_per_user` entry

# WHEN arg is -1
# THEN rendered config does not have a `max_global_exemplars_per_user` entry

# WHEN arg is 1
# THEN rendered config has `max_global_exemplars_per_user: 1` entry
```

charmcraft.yaml
```diff
@@ -150,3 +150,9 @@ requires:
     optional: true
     description: |
       Integration to help users discover Mimir's deployment.
+config:
+  options:
+    max_global_exemplars_per_user:
+      default: 100000
+      description: The limit for exemplars
```
sed-i, line 157: How about (please confirm):
```suggestion
      description: |
        The global limit for number of exemplars.
        When unset or set to a non-positive value, exemplar storage is disabled.
```

src/charm.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, *args: Any):
             scheme=lambda: urlparse(self.internal_url).scheme,
         )
         self.alertmanager = AlertmanagerConsumer(charm=self, relation_name="alertmanager")
+        self.max_global_exemplars_per_user = self.config['max_global_exemplars_per_user']
```
sed-i, line 73: Do we need to keep this value in `self`? Seems like it is used only once below.

How about the following tests:

```python
# WHEN config value is unset
# THEN outgoing relation data to worker has `max_global_exemplars_per_user: 100000`

# WHEN config value changes from unset to 0
# THEN outgoing relation data to worker does not have `max_global_exemplars_per_user`
```

sed-i (commented):
Nice! Some comments :)

charmcraft.yaml
```diff
@@ -150,3 +150,13 @@ requires:
...
       Integration to help users discover Mimir's deployment.
+config:
+  options:
+    max_global_exemplars_per_user:
+      default: 0
+      description: |
+       The global limit for the number of exemplars.
+       When unset or set to a non-positive number, exemplar storage is disabled.
+       If set to a positive number that falls between 1 and 100,000, the limit is bumped to 100,000.
+       To disable, set to a non-positive number.
```
lucabello, line 161: I would remove this line: to disable, they should either set to 0 or unset; you can probably remove this entirely.

tests/unit/test_charm.py
```diff
@@ -52,3 +52,33 @@ def test_recommended(mock_coordinator, roles, expected):
...
+        elif 1 <= max_global_exemplars_per_user <= 100000:
+            return 100000
+        else:
+            return max_global_exemplars_per_user
+
+    mc._workers_config_getter = MagicMock(side_effect=side_effect)
+
+    result = mc._workers_config_getter()
+
+    assert result is expected_value
```
sed-i, line 84: This test is very similar to the one in test_mimir_config.
And it's a whitebox test.
In charm unit tests it's more beneficial to to "zoom out" a bit. For example:

```python
# GIVEN the config option is unset
# WHEN a worker relation joins
# THEN relation data has "0"

# GIVEN the config option is 99999
# WHEN a worker relation joins
# THEN relation data has "100000"

# GIVEN the config option is 10001
# WHEN a worker relation joins
# THEN relation data has "10001"
```

tests/unit/helpers.py
```diff
@@ -1,5 +1,27 @@
...
+def get_worker_config_exemplars(relations, endpoint):
+    # Find the relevant relation
+    relevant = [r.local_app_data for r in relations if r.endpoint == endpoint]
+
+    assert relevant, "No matching relation found"
+
+    worker_config_str = relevant[0]['worker_config']
+
+    worker_config = yaml.safe_load(worker_config_str)
+    worker_config = yaml.safe_load(worker_config)
```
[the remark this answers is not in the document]
lucabello: I wouldn't expect we need to `yaml.safe_load` twice

MichaelThamm:
Consider looking into creating a feature PR in avalanche to enable trace IDs:
- https://github.com/prometheus-community/avalanche

Consider using telemetrygen from Opentelemetry:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/telemetrygen/README.md

tests/unit/helpers.py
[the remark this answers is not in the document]
sinapah: I think we're dealing with YAML that contains YAML( a YAML string inside a YAML doc).

The `worker-config-str` is the inner YAML which is serialized and stored as a string value in the outer YAML. 

The first `yaml.safe_load()` parses the outer YAML.

It returns a string that looks like YAML, but is still a string .

The second `yaml.safe_load()` parses that string into an actual Python dict.

MichaelThamm (changes requested):

tests/integration/helpers.py
```diff
@@ -146,3 +152,75 @@ async def get_traefik_proxied_endpoints(
...
+        with tracer_provider.get_tracer("service").start_as_current_span("generate_metrics_span") as span:
+            # Extract trace_id from the current span
+            span_ctx = span.get_span_context()
+            trace_id = span_ctx.trace_id
+
+            # Now, we have to convert the decimal trace ID above into Hex because when querying the exemplars, the trace ID returned will be base 16
+            trace_id_hex = format_trace_id(trace_id)
+
+            # Record a random value for the counter and include the trace_id as part of the exemplar
+            counter.add(random.randint(1, 10), {"trace_id":trace_id})
```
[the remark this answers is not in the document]
sinapah: You're right, no need to have the random value. We can just assign an arbitrary int there

tests/integration/helpers.py
```diff
@@ -146,3 +151,78 @@ async def get_traefik_proxied_endpoints(
...
+
+    # Create a Meter instance
+    meter = metrics.get_meter("meter", "1.0.0")
+    counter = meter.create_counter(metric_name, description="A placeholder counter metric")
+
+    # Set up tracing (TracerProvider) to generate trace_id
+    tracer_provider = TracerProvider()
+
+    # Function to simulate metric data and include exemplar (trace_id)
+    def generate_and_record_metrics():
```
MichaelThamm, line 183: Consider pulling this function outside so it is not nested.

MichaelThamm (changes requested):
Please address the comments and fix the CI, otherwise LGTM!

src/mimir_config.py
```diff
@@ -314,7 +318,16 @@ def _build_memberlist_config(
...
     # Maximum number of rule groups per-tenant. 0 to disable. (default 70)
     def _build_limits_config(self) -> Dict[str, Any]:
-        return {
+        limits_config: Dict[str, Any] = {
             "ruler_max_rules_per_rule_group": 0,
             "ruler_max_rule_groups_per_tenant": 0,
         }
+
+        # Set the max_global_exemplars_per_user based on the value of _max_global_exemplars_per_user
+        if val := max(self._max_global_exemplars_per_user or 0, 0):
```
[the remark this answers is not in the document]
sinapah: Yes, this is to protect against negative values. There is a small scenario where `self._max_global_exemplars_per_user or 0` falls back to 0 and that's when the variable is negative, i.e. the user set it to negative in the config. In that case, we need to fallback to 0.
When the value is negative, inn `val = max(max(self._max_global_exemplars_per_user, 0), EXEMPLARS_FLOOR)`, the inner max will evaluate 0 and the second one will evaluate to the EXEMPLARS_FLOOR (100_000), which goes against the desired behaviour where we need it to evaluate to 0.

You're right that None or False are not allowed.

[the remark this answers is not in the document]
MichaelThamm: Oh I see, we only want to use floor when the user sets a valid value, otherwise disabled. I did read this in your config description. LGTM

MichaelThamm (approved):
LGTM. You still have some conversations to resolve, but I am happy with the changes.

PR #156, opened 2025-06-27
feat: utests for ingress TLS

MichaelThamm:
## Issue

Similar to:
- https://github.com/canonical/loki-coordinator-k8s-operator/pull/69

## Solution

Add unit tests for validating databag contents over `ingress` relation contains correct information when TLS is enabled.

## Context

This PR includes some itest updates to bump our charms to deploy from `2/edge`

## Testing Instructions

`tox -e unit`

### DIFF
--- a/tests/integration/test_charm_ha.py
+++ b/tests/integration/test_charm_ha.py
@@ -9,7 +9,6 @@ import logging
 import pytest
 import requests
 from helpers import (
-    COS_CHANNEL,
     charm_resources,
     configure_minio,
     configure_s3_integrator,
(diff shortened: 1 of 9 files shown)


tests/integration/test_charm_ha.py
```diff
@@ -9,6 +9,7 @@
 import pytest
 import requests
 from helpers import (
+    COS_CHANNEL,
```
sed-i, line 12: This is fine, but perhaps would make more sense as a fixture? Not sure.

tests/unit/conftest.py
```diff
@@ -52,11 +52,21 @@ def nginx_container():
...
+            Exec(
+                [
+                    "mimirtool",
+                    "rules",
+                    "sync",
+                    f"--address=https://{socket.getfqdn()}:{NGINX_TLS_PORT}",
+                    "--id=anonymous",
+                ],
+                return_code=0,
+            ),
```
sed-i, line 69: Nit: if you store the addresses in vars, e.g. `http_arg = f"--addres=..."` then perhaps the code will fold into something nicer.

tests/unit/helpers.py
```diff
@@ -0,0 +1,3 @@
+def get_relation_data(relations, endpoint, key):
+    """Retrieve the value for a given key from the local_app_data of a relation with the specified endpoint."""
+    return next((r.local_app_data[key] for r in relations if r.endpoint == endpoint), None)
```
sed-i, line 3: To avoid surprises in the future, we could address a potential future case of more than one relation of the same type by adding an assertion.


```suggestion
    relevant = [r.local_app_data[key] for r in relations if r.endpoint == endpoint]
    assert len(relevant) < 2, "This helper currently assumes only one relation."
    return relevant[0] if relevant else None
```

PR #178, opened 2025-09-22
feat: add retention period charm config

### LINKED ISSUE #150: Expose retention period as charm configuration
### Enhancement Proposal

Similar to the Loki and Tempo charms, the retention period passed to the underlying Mimir app should be exposed as a charm configuration option. This option would help retain only relevant and recent metrics.

The underlying app supports this option: https://grafana.com/docs/mimir/latest/configure/configure-metrics-storage-retention/ .

sinapah:
## Issue

Fixes #150.

## Solution

Adds a Juju config option for setting the retention period (called `metrics_retention_period` for consistency with Prometheus). Valid units are y, m, w, d, s, ms. "0" disabled data deletion.
When the user enters an invalid option, the charm is blocked and data deletion is suspended by setting the worker's retention to 0. This is to avoid data loss.
Ref: https://grafana.com/docs/mimir/latest/configure/configure-metrics-storage-retention/

Furthermore, [a recent PR](https://github.com/canonical/cos-lib/pull/154) in `cosl` added a timespec validation check which was released in version 1.2.0. This PR bumps the version of `cosl` to 1.2.0 and imports the check.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -160,7 +160,7 @@ config:
         Otherwise, the value is set to the greater of the setpoint or 100,000.
         Ref: https://grafana.com/docs/mimir/latest/manage/use-exemplars/store-exemplars/
       type: int
-    blocks_retention_period:
+    metrics_retention_period:
       default: "0"
       description: |
         Sets a global retention period for metrics in Mimir. Supported units: d, w, m, y, h, s, ms or 0 for no limit (disable)
--- a/src/charm.py
+++ b/src/charm.py
@@ -65,7 +65,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
             scheme=lambda: urlparse(self.internal_url).scheme,
         )
         self.alertmanager = AlertmanagerConsumer(charm=self, relation_name="alertmanager")
-        self.retention_period = str(self.config['blocks_retention_period'])
+        self.retention_period = str(self.config['metrics_retention_period'])
         self.coordinator = Coordinator(
             charm=self,
             roles_config=MIMIR_ROLES_CONFIG,
@@ -95,7 +95,7 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
                 topology=JujuTopology.from_charm(self),
                 alertmanager_urls=self.alertmanager.get_cluster_info(),
                 max_global_exemplars_per_user=int(self.config["max_global_exemplars_per_user"]),
-                blocks_retention_period=self.retention_period if is_valid_timespec(self.retention_period) else "0",
+                metrics_retention_period=self.retention_period if is_valid_timespec(self.retention_period) else None
             ).config,
             worker_ports=lambda _: tuple({8080, 9095}),
             resources_requests=self.get_resource_requests,
@@ -371,18 +371,11 @@ class MimirCoordinatorK8SOperatorCharm(ops.CharmBase):
         """Returns a dictionary for the "requests" portion of the resources requirements."""
         return {"cpu": "50m", "memory": "100Mi"}
 
-    def _truncate_string_if_too_long(self, string: str) -> str:
-        """Returns a string unchanged if it's under 5 characters or truncates it using ... if 5 characters or more."""
-        max_length = 4
-        if len(string) <= max_length:
-            return string
-        return f"{string[:max_length]}..."
-
     def _on_collect_unit_status(self, event: ops.CollectStatusEvent):
         event.add_status(ActiveStatus())
         if not is_valid_timespec(self.retention_period):
             logger.info(f"Suspending data deletion due to invalid option set in config: {self.retention_period}. To resume data deletion, please reset value to a valid option.")
-            event.add_status(BlockedStatus(f"Invalid config option: retention_period={self._truncate_string_if_too_long(self.retention_period)}; see debug-log"))
+            event.add_status(BlockedStatus(f"Invalid config option (see debug-log): retention_period={self.retention_period}"))
 
     def _reconcile(self):
         # This method contains unconditional update logic, i.e. logic that should be executed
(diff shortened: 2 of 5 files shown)


charmcraft.yaml
```diff
@@ -155,8 +155,15 @@ config:
...
-       Otherwise, the value is set to the greater of the setpoint or 100,000.
-       Ref: https://grafana.com/docs/mimir/latest/manage/use-exemplars/store-exemplars/
-      type: int
\ No newline at end of file
+        The global limit for the number of exemplars.
+        When unset or set to a non-positive number, exemplar storage is disabled.
+        Otherwise, the value is set to the greater of the setpoint or 100,000.
+        Ref: https://grafana.com/docs/mimir/latest/manage/use-exemplars/store-exemplars/
+      type: int
+    blocks_retention_period:
```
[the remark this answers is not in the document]
sinapah: According to [the upstream docs](https://grafana.com/docs/mimir/latest/configure/configure-metrics-storage-retention/#configure-the-storage-retention):

> The [compactor](https://grafana.com/docs/mimir/latest/references/architecture/components/compactor/) is the Mimir component that is responsible for enforcing the storage retention.

Thus, the compactor is the only component related to storage retention. With that said, I think leaving the prefix `compactor_` should be fine, as it keeps the config value shorter. I would even suggest changing it to `retention_period` or `metrics_retention_period` for consistency with Prometheus, but I doubt we need `compactor_` as a prefix here.

src/charm.py
```diff
@@ -367,6 +371,19 @@ def get_resource_requests(self, _) -> Dict[str, str]:
         """Returns a dictionary for the "requests" portion of the resources requirements."""
         return {"cpu": "50m", "memory": "100Mi"}
 
+    def _truncate_string_if_too_long(self, string: str) -> str:
+        """Returns a string unchanged if it's under 5 characters or truncates it using ... if 5 characters or more."""
+        max_length = 4
+        if len(string) <= max_length:
+            return string
+        return f"{string[:max_length]}..."
```
lucabello, line 379: non-blocking: I don't think we need this function, because the "too long" string only happens in case of misconfiguration, such as

```bash
juju config mimir blocks_retention_period="someVeryReallySuperLongStringThatIsWrong"
```

src/charm.py
```diff
@@ -63,6 +65,7 @@ def __init__(self, *args: Any):
             scheme=lambda: urlparse(self.internal_url).scheme,
         )
         self.alertmanager = AlertmanagerConsumer(charm=self, relation_name="alertmanager")
+        self.retention_period = str(self.config['blocks_retention_period'])
```
lucabello, line 68:
```suggestion
		retention_period_raw = str(self.config['blocks_retention_period'])
        self.retention_period: Optional[str] = retention_period_raw if is_valid_timespec(retention_period_raw) else None
```

src/charm.py
```diff
@@ -92,6 +95,7 @@ def __init__(self, *args: Any):
                 topology=JujuTopology.from_charm(self),
                 alertmanager_urls=self.alertmanager.get_cluster_info(),
                 max_global_exemplars_per_user=int(self.config["max_global_exemplars_per_user"]),
+                blocks_retention_period=self.retention_period if is_valid_timespec(self.retention_period) else "0",
```
lucabello, line 98:
```suggestion
                blocks_retention_period=self.retention_period,
```

src/mimir_config.py
```diff
@@ -121,12 +121,14 @@ def __init__(
...
         root_data_dir: Path = Path("/data"),
         recovery_data_dir: Path = Path("/recovery-data"),
+        blocks_retention_period: Optional[str] = None,
     ):
         self._alertmanager_urls = alertmanager_urls
         self._root_data_dir = root_data_dir
         self._recovery_data_dir = recovery_data_dir
         self._max_global_exemplars_per_user = max_global_exemplars_per_user
         self._topology = topology
+        self._blocks_retention_period = blocks_retention_period
```
lucabello, line 131:
```suggestion
        self._blocks_retention_period: str = blocks_retention_period or "0"
```

src/mimir_config.py
```diff
@@ -351,4 +353,11 @@ def _build_limits_config(self) -> Dict[str, Any]:
             val = max(val, EXEMPLARS_FLOOR)
         limits_config["max_global_exemplars_per_user"] = val
 
+        # If the config value is invalid, the charm will pass the string "0" to MimirConfig.
+        # We'll turn that into an int so it shows up like compactor_blocks_retention_period: 0 in the worker's config YAML file
+        # And not compactor_blocks_retention_period: '0'. Both are valid, but the Grafana docs use 0 (https://grafana.com/docs/mimir/latest/configure/configure-metrics-storage-retention/).
+        # This is for consistency.
+        limits_config["compactor_blocks_retention_period"] = int(self._blocks_retention_period) if self._blocks_retention_period == "0" else self._blocks_retention_period
```
lucabello, line 360:
```suggestion
        limits_config["compactor_blocks_retention_period"] = 0 if self._blocks_retention_period == "0" else self._blocks_retention_period
```

lucabello (approved):
Great work! Approving with some comments :)

src/charm.py
sed-i: I agree that it's odd to have a one-off truncation just in this charm here, but the [50 chars limit from juju](https://github.com/juju/juju/pull/14867) applies to the entire message; here we want to make sure "see debug-log" is printed.

Perhaps if we change the message structure a bit then we won't need this function:
```python
f"Invalid config option (see debug-log): retention_period={...}"
```

Apologies for the back and forth given I asked for the original impl 😅

sed-i (approved):

src/charm.py
sinapah: Thanks for the feedback, it has been removed :slightly_smiling_face:

